PR #26, opened 2015-01-28
File persistence

jbout:


llecaque:
getModelIdFromXml is in both FileIterator and FileModel

PR #36, opened 2015-02-17
Allow permision providers to be configurable using Configurable

jbout:


jbout:
Depends on https://github.com/oat-sa/extension-tao-dac-simple/pull/3

PR #39, opened 2015-03-03
Refactor/smooth configurable

jbout:
Moved the configuration of the smooth model from the singletons to the ontology configuration


jbout:
Please take a look @bugalood

PR #40, opened 2015-03-04
add the name on lastInsertId call

antoinerobin:
hotfix for postgresql driver on lastInsertId method


bugalood:
I take the entire/full/absolute responsibility of this :+1:

PR #57, opened 2015-04-10
Added checking for constraint matching during setting passwords

boajer:


bugalot:
Thanks a lot for your efforts, and sorry for this loooooong time delay!

PR #70, opened 2015-07-12
Tao 987 couchbase persistence implementation

bursonic:


jbout:
Split out into https://github.com/oat-sa/lib-advkv-couchbase

PR #73, opened 2015-07-17
fixed INSERT query builder. used to fail on some queries on PgSQL

bursonic:
Related: oat-sa/extension-tao-revision#29


jbout:
duplicate of https://github.com/oat-sa/generis/pull/74/files

PR #82, opened 2015-08-10
Added Report::createInfo and Report::getSuccesses convenience methods.

bugalot:
Removed @deprecated as discussed offline.

### DIFF
--- a/common/report/class.Report.php
+++ b/common/report/class.Report.php
@@ -65,9 +65,9 @@ class common_report_Report implements IteratorAggregate, JsonSerializable
     private $data = null;
 	
 	/**
-	 * convenience methode to create a simple success report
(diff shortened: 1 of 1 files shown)


common/report/class.Report.php
```diff
@@ -175,9 +185,22 @@ public function getSuccesses() {
...
+	    $infos = array();
+	    foreach ($this as $element) {
+	        if ($element->getType() == self::TYPE_INFO) {
+	            $infos[] = $element;
+	        }
+	    }
+	    return $infos;
+	}
+	
+	/**
```
antoinerobin: wouldn't it be better to have a 
```php
private function getElements($type)
````
And then call it in getSuccesses, getErrors and getInfos ?

PR #93, opened 2015-08-27
normalize paths

hutnikau:


jbout:
parameters should be corrected before calling the function else we can no longer differentiate between:

a / b\c / d
and
a / b / c / d

PR #94, opened 2015-09-03
TAO 1227:  session uses User interface now instead of abstract user class

bursonic:
Master PR: oat-sa/extension-tao-delivery#100


antoinerobin:
please see my comment in https://github.com/oat-sa/extension-tao-delivery/pull/100
You have some modifications to do about interfaces

bursonic:
Updated: new Refreshable interface added to respect persisted User abstract class signature

PR #99, opened 2015-09-28
Returns errors only for invalid sub validators

boajer:


core/user/PasswordConstraintsService.php
```diff
@@ -124,14 +124,19 @@ protected function register($config)
...
      * @return array
      */
-    public function getErrors()
+    public function getErrors($password = '')
     {
         $errors = array();
         /** @var \tao_helpers_form_Validator $validator */
         foreach ($this->validators as $validator) {
-            $errors[] = $validator->getMessage();
+            if ( ! $validator->evaluate($password)) {
```
antoinerobin: it may be better to fill the error array in the validate method. So we have a method to validate, that returns true or false and another that returns errors without having to re-validate

helpers/class.PasswordHash.php
```diff
@@ -26,15 +26,17 @@ public function __construct($algorithm, $saltLength) {
...
+        if (0 === count($errors)) {
             $salt = helpers_Random::generateString($this->saltLength);
             return $salt.hash($this->algorithm, $salt.$password);
         }
 
-        throw new PasswordConstraintsException(
-            __( 'Password must be: %s' ,
-            implode( ',', PasswordConstraintsService::singleton()->getErrors() )
-        ));
+        $exception = new PasswordConstraintsException('Password must be: %s' . implode(',', $errors));
```
antoinerobin: the %s doesn't work for me I've got ```Password must be: %sThis field is too short (minimum 15)```

core/user/PasswordConstraintsException.php
```diff
@@ -22,4 +22,17 @@
...
+     * @param array $errors
+     *
+     * @return PasswordConstraintsException
+     */
+    public function setErrors(array $errors)
+    {
+        $this->errors = $errors;
+
+        return $this;
+    }
```
antoinerobin: Is this really needed ? I can't find where we use these errors

PR #102, opened 2015-10-14
Initial event manager commit

jbout:
Very simple event manager


common/oatbox/event/DoctrineEventManager.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    private $eventManager;
+    
+    public function trigger($event, $params = array()) {
+        $eventObject = is_object($event) ? $event : new GenericEvent($event);
+        $eventArgs = new EventArgs();
+        $this->getRealEventManager()->dispatchEvent($eventObject->getName(), $eventArgs);
+    }
+    
+    public function attach($event, $callback) {
```
Alroniks: Not important, but it would be good check callable type for callback for prevent weird behavior in future.
```php
public function attach($event, Callable $callback) {}
```

jbout: That class was experimental and should not be part of it

PR #103, opened 2015-10-16
base implementation with messages

krampstudio:
Seeing server side logging in the browser dev tools [Native in Firefox 43](https://developer.mozilla.org/en-US/docs/Tools/Web_Console/Console_messages#Server) or using an [extension in Chromium](https://chrome.google.com/webstore/detail/chrome-logger/noaneddfkdjfnfdakjjmocngnfkfehhd)


krampstudio:
Moving to taoDevTools

PR #104, opened 2015-10-20
use the platform limit statement

antoinerobin:


jbout:
Breaks test cases

Prophecy\Exception\Call\UnexpectedCallException: Method call:
  - getPlatForm()

PR #105, opened 2015-10-21
Allow multi level config entries

jbout:


jbout:
headers and default values not yet handled correctly for nested configs

PR #122, opened 2015-11-16
Refactored repository class, add ability define callback for set cust…

Alroniks:
…om filename

Related to https://oat-sa.atlassian.net/browse/TAO-1202
https://github.com/oat-sa/extension-tao-revision/pull/33

### DIFF
--- a/core/kernel/versioning/class.Repository.php
+++ b/core/kernel/versioning/class.Repository.php
@@ -127,13 +127,7 @@ class core_kernel_versioning_Repository
      */
     public function getPath()
     {
-        $returnValue = (string) '';
-
-        
-        $returnValue = core_kernel_fileSystem_Cache::getFileSystemPath($this);
-        
-
-        return (string) $returnValue;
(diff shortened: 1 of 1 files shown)


core/kernel/versioning/class.Repository.php
```diff
@@ -400,23 +394,30 @@ public function getRootFile()
      * @author Jerome Bogaerts, <jerome@taotesting.com>
      * @param  string filePath The path to the file you want the repository to deal with.
      * @param  string label A label for the created file Resource.
+     * @param  callable $createFileName A function that generates unique file name during spawn file
      * @return core_kernel_versioning_File
      * @since 2.4
      */
-    public function spawnFile($filePath, $label = '')
+    public function spawnFile($filePath, $label = '', callable $createFileName = null)
```
antoinerobin: Why do you have to have a callable to create the filename ? the one here is enough, no ?

Alroniks: Default method hardcoded to use crc32 function, but webservices in taoQtiTest use md5. I think it's good to have ability to customise this behavior.

PR #152, opened 2016-02-11
Allow alternate config peristence

jbout:
@llecaque can you take a look as well please?


hutnikau:
Is this PR still relevant?
If so, could you please remind me why we need this change?

jbout:
Afair, this was intended to be able to unit test the registry

jbout:
outdated

PR #154, opened 2016-02-22
LoginFailedException::getUserMessage() returns an incorrect result

zShurick:


jbout:
Exception should never be a string

PR #161, opened 2016-03-08
Add Lower() function to search pattern

hutnikau:


jbout:
Execution time for search queries is only slightly increased for MySql, but impact on Postgres is noticable:
MySql: +2%
Postgresql: +25%

PR #172, opened 2016-03-30
FileSystem: Fixed php fatal error, when location is not readable

zagovorichev:


jbout:
While edge cases still exist, this covers most issues of the past

PR #173, opened 2016-03-30
Fixed fatal error, when have not rights to the attached files

zagovorichev:


common/configuration/class.FileSystemComponent.php
```diff
@@ -244,38 +244,54 @@ public function check()
...
-    /**
-     * If file is readable.
-     *
-     * @access public
-     * @author Aleh Hutnikau, <hutnikau@1pt.com>
-     * @param string $location File location
-     * @return boolean
-     */
-    public function isReadable($location = null)
+    private function hasLocationAccess($location = null, $rule = 'Readable')
```
siwane: Could you check if $rule parameter is valid. 

> in_array($rule, array('writable', 'readable', 'executable'))

And with PHP doc block, it should be perfect

zagovorichev: It's a private method, all rules are defined in this class as static string
What can I add to Doc? For me it's clear from function name

siwane: Ok, I agree with your point of view about PHP doc block

PR #177, opened 2016-04-19
get list of actions

hutnikau:


jbout:
Fixed underlying issue instead of blacklisting: https://github.com/oat-sa/extension-tao-outcomeui/pull/63

jbout:
Missing Upgrade script & version increase

PR #180, opened 2016-04-25
Fixed  password rule for uppercase letters

zagovorichev:


core/user/PasswordConstraintsService.php
```diff
@@ -88,7 +88,7 @@ protected function register($config)
             $this->validators[] = new \tao_helpers_form_validators_Regex(
                 array(
                     'message' => __( 'Must include upper case letters' ),
-                    'format'  => '/(\p{Ll}+.*\p{Lu})|(\p{Lu}+.*\p{Ll})/'
+                    'format'  => '/(\p{Lu}+)/',
```
boajer: According old rule password has been matched against regex 'mix of lowercase and uppercase' and indeed message is not accurate. But after your changes it's impossible to turn on simultaneously 'upcase' and 'lowcase' rules in config. 
I think it's better to adjust message itself or change logic and update configs for this feature

PR #193, opened 2016-06-17
Fix/sql exposing

boajer:
Prevents exposing sql queries into responses to XMLHttpRequest


boajer:
Different approach has been chosen.

PR #200, opened 2016-06-28
Added first PSR 3 logger trait

jbout:


common/oatbox/log/LoggerAggregator.php
```diff
@@ -0,0 +1,48 @@
...
+class LoggerAggregator extends AbstractLogger
+{
+    /**
+     * @var LoggerInterface[]
+     */
+    private $loggers;
+    
+    public function __construct($loggers)
+    {
+        $this->loggers = $loggers;
```
siwane: Is it a good idea to verify here that instance of $loggers element is _LoggerInterface_ ?
I see that this check is done into LoggerAwareTrait

jbout:
remerged develop

PR #208, opened 2016-07-25
Use provided config service in service manager

jbout:


bugalot:
RAS in the context of the CTA project.

PR #224, opened 2016-08-22
Cache extension dependencies

hutnikau:
During profiling of call to `/taoItems/Items/editItemClass` action I found that the most of the time takes initialization of tao. I found out that `common_ext_Extension::getDependencies()` and `common_ext_Manifest::getDependencies()` was called 1574 times

Ideally all those function should be called no more times than amount of extensions in tao. I made small improvements to cache results of those functions. I got the following results:
getDependencies(): 119 vs 1574 calls

Time of execution of  `/taoItems/Items/editItemClass` request was reduced on 20% (119ms vs 143ms)

Not really big acceleration but i guess it worth to merge.

### DIFF
--- a/common/ext/class.Extension.php
+++ b/common/ext/class.Extension.php
@@ -39,6 +39,20 @@ class common_ext_Extension
      */
     const MANIFEST_NAME = 'manifest.php';
 
+    /**
+     * Cache storage
+     *
+     * @var string
+     */
+    const STORAGE = 'cache';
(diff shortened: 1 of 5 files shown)


hutnikau:
I have resolved conflicts. Please review it again.

bugalot:
@jbout , as I will be out for two weeks, you are our only hope ! :+1: on this PR by the way

PR #228, opened 2016-08-26
Fixed exceptions type

boajer:


jbout:
The current exception is indeed not correct, but exposing the Flysystem exception might be problematic as well.

I would suggest to create a dedicated exception for this case and return it, independent of the implementation.

PR #231, opened 2016-09-02
Fix/incorrect event detaching

boajer:
In case of event have several listeners attached and one of them ( not last one ) is going to be detached, array turns in associative with numeric keys and gap in sequence see bellow

`'event' => array(
            0 => $callback1,
            2 => $callback2,
            3 => $callback3
)
`

Afterwards attempt to detach any of listeners with index more than previously removed will fail


hutnikau:
I guess that we must avoid of appearing keys in the listeners array. Maybe better option is use `array_values()` for listeners array after detaching?
```php
$listeners[$eventObject->getName()] = array_values($listeners[$eventObject->getName()]);
$this->setOption(self::OPTION_LISTENERS, $listeners);
```
@jbout what do you think about this?

PR #243, opened 2016-10-03
namespaced install/uninstall helper

chrisphpgarcia:


common/ext/action/class.InstallAction.php
```diff
@@ -25,6 +25,7 @@
 /**
  * Abstract action containing some helper functions
  * @author bout
+ * @deprecated since version 3.2.0 please use oat\oatbox\extension\Installer instead
  *
  */
 abstract class common_ext_action_InstallAction implements Action, ServiceLocatorAwareInterface
```
jbout: extend the new InstallAction instead of having duplicate code?

chrisphpgarcia: because I won't to have installation methods in Uninstaller.

PR #244, opened 2016-10-04
Remove redundant union query

hutnikau:
`searchInstances` returns absolutely wrong result (it does not take in account any options such as filters and so on).
To reproduce this bug you have to create sample resources:
https://gist.github.com/hutnikau/04d051fb446d7fadeee543e86101a078
and run search test:
https://gist.github.com/hutnikau/048466408179f21916a92a12f5077766

You will see that result will has much more instances than expected (8 insead 3 for me).

Sql Query will be:
```sql
SELECT "subject" FROM ( 
  (SELECT DISTINCT "subject" FROM "statements" WHERE 
	 "predicate" = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type' AND  
	 ("object" IN ('http://www.tao.lu/Ontologies/TAOSubject.rdf#Subject' , 'http://www.tao.lu/Ontologies/TAO.rdf#User')  )
  )  
  UNION ALL   
  (SELECT DISTINCT "subject" FROM "statements" 
    WHERE  "predicate" = 'http://www.tao.lu/Ontologies/TAOTestCenter.rdf#authorizedProctor' AND 
    ("object" = 'http://myTest.case#TestCenter1' 
    OR  "object" = 'http://myTest.case#TestCenter2'  
    OR  "object" = 'http://myTest.case#TestCenter3')) 
) AS unionq GROUP BY "subject" HAVING COUNT(*) >=2 
UNION SELECT "subject" FROM (  
  (SELECT DISTINCT "subject" FROM "statements" WHERE  
  "predicate" = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type' AND 
  ("object" IN ('http://www.tao.lu/Ontologies/TAOSubject.rdf#Subject' , 'http://www.tao.lu/Ontologies/TAO.rdf#User')  ))
) AS unionq GROUP BY "subject" HAVING COUNT(*) >=1 
```
the last SELECT is redundant there.
I tried to find the reason of this additional query but did not succeed.

### DIFF
--- a/core/kernel/persistence/smoothsql/search/ComplexSearchService.php
+++ b/core/kernel/persistence/smoothsql/search/ComplexSearchService.php
@@ -185,36 +185,50 @@ class ComplexSearchService extends ConfigurableService
      */
     public function getQuery(core_kernel_persistence_smoothsql_SmoothModel $model, $classUri, array $propertyFilters, $and = true, $like = true, $lang = '', $offset = 0, $limit = 0, $order = '', $orderDir = 'ASC') 
     {
-        $query = $this->gateway->query()->setLimit($limit)->setOffset($offset);
+        $query = $this->getGateway()->query()->setLimit( $limit )->setOffset($offset );
         
-        if (!empty($order)) {
+        if(!empty($order)) {
             $query->sort([$order => strtolower($orderDir)]);
         }
         
-        $this->setLanguage($query, $lang);
-
+        $this->setLanguage($query , $lang);
+        
+        $operator = $this->getOperator($like);
+        
         $criteria = $query->newQuery()
-            ->add('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')
-            ->in($classUri);
-
-        $queryBuilder = $query->setCriteria($criteria);
-
-        $count = 0;
-        foreach ($propertyFilters as $predicate => $value) {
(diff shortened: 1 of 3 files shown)


core/kernel/persistence/smoothsql/search/ComplexSearchService.php
```diff
@@ -211,12 +211,6 @@ public function getQuery(core_kernel_persistence_smoothsql_SmoothModel $model, $
             foreach ($nextValue as $val) {
                 $criteria->addOr($this->parseValue($val));
             }
-            if($and === false) {
```
chrisphpgarcia: It mustn't be added if it the last criterion.

hutnikau: @chrisphpgarcia but this condition added twice:
https://github.com/oat-sa/generis/blob/develop/core/kernel/persistence/smoothsql/search/ComplexSearchService.php#L193-L194
and
https://github.com/oat-sa/generis/blob/develop/core/kernel/persistence/smoothsql/search/ComplexSearchService.php#L216-L217

hutnikau:
@chrisphpgarcia please look at this PR one more time. I have refactored this method and now it seems like `searchInstances()` works fine with OR and AND conditions.

hutnikau:
This approach does not work. All the filters on grids are broken (for example users grid on Manage Users page).

PR #256, opened 2016-10-20
check HTTP_X_FORWARDED_PROTO

hutnikau:


bugalot:
Just for PR reader's comprehension:

- http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-proto
- http://stackoverflow.com/questions/16042647/whats-the-de-facto-standard-for-a-reverse-proxy-to-tell-the-backend-ssl-is-used

PR #258, opened 2016-10-21
Feature/get sub service

jbout:

PR #260, opened 2016-10-21
Recreated DeDuplication service branch

jbout:
requires:
* https://github.com/oat-sa/generis/pull/258
* https://github.com/oat-sa/generis/pull/259


core/resource/AbstractCreateOrReuse.php
```diff
@@ -0,0 +1,141 @@
...
+        $gateWay       = $searchService->getGateway();
+        
+        $searchQueryBuilder = $gateWay->query();
+        
+        $searchService->searchType($searchQueryBuilder, $this->getParentClass()->getUri() , true);
+        
+        $criterion = $searchQueryBuilder->newQuery();
+        
+        foreach ($this->getUniquePredicates() as $field) {
+            $value = $values[$field];
```
siwane: Key exist not checked

core/resource/AbstractCreateOrReuse.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+    
+    /**
+     * 
+     * @param array $values
+     * @return TaoResultSet
+     */
+    protected function searchResource(array $values) {
+        
+        $searchService = $this->getSearchService();
```
siwane: You should use Model abstraction and not directly call search service (smoothmodel)

PR #266, opened 2016-11-07
fix ontology update if workspace was installed

chrisphpgarcia:


scripts/update/Updater.php
```diff
@@ -283,8 +288,24 @@ public function update($initialVersion) {
...
+                $inner->setOption(\core_kernel_persistence_smoothsql_SmoothModel::OPTION_SEARCH_SERVICE , ComplexSearchService::SERVICE_ID);
+
+                $workspace = $modelWrapper->getWorkspaceModel();
+                $workspace->setOption(core_kernel_persistence_smoothsql_SmoothModel::OPTION_SEARCH_SERVICE , ComplexSearchService::SERVICE_ID);
 
-        $this->skip('3.7.0', '3.8.3');
+                $wrapedModel = WrapperModel::wrap($inner, $workspace );
+                $wrapedModel->setServiceLocator($this->getServiceManager());
+                ModelManager::setModel($wrapedModel);
+            }
```
antoinerobin: add a else at least to know that nothing has to be done

PR #289, opened 2017-01-10
Use SERVICE_ID instead of CONFIG_ID

jbout:


chrisphpgarcia:
this PR is configured to merger in master. Is that normal ?

PR #296, opened 2017-01-24
get queue  result properly

boajer:


core/kernel/persistence/smoothsql/search/GateWay.php
```diff
@@ -119,12 +119,13 @@ public function fetchQuery($query) {
     /**
      * return total count result
      * @param QueryBuilderInterface $Builder
-     * @return type
+     * @return integer
      */
     public function count(QueryBuilderInterface $Builder) {
         $this->parsedQuery = $this->getSerialyser()->setCriteriaList($Builder)->count(true)->serialyse();
```
hutnikau: What is the reason to use `$this` here? `parsedQuery` property is used somewhere else?

PR #300, opened 2017-02-01
Fix/error msg extension

jbout:
Better error reporting of unsatisfied requirements


gyszucs:
Is there a way to have this exception thrown somehow?

jbout:
Add an extension which doesn't exist as required to one of your extensions

siwane:
what do u think about auto install if the extension exists but not installed?

jbout:
I worry that it would be "too magically" , especially in an AWS environment, where one of the web-servers would suddenly automatically install an extension

helpers/class.ExtensionHelper.php
```diff
@@ -69,7 +69,11 @@ public static function sortByDependencies($extensions) {
                 }
             }
             if (count($unsorted) == $before) {
-                throw new common_exception_Error('Unable to resolve extension dependencies');
+                if (count($unsorted) == 1) {
+                    throw new common_exception_Error('Missing extensions '.implode(',', $missing).' for: '.implode(',', array_keys($unsorted)));
```
siwane: Do u have a more elegant way to do that? ( ```if (count($unsorted) == 1 : implode(',', array_keys($unsorted)));```)

PR #301, opened 2017-02-01
TAO-3698 Support permision services

jbout:


core/data/permission/PermissionManager.php
```diff
@@ -35,29 +37,22 @@ class PermissionManager
...
         }
-        return self::$model;
     }
     
     /**
      * @param core_kernel_persistence_RdfsDriver $model
      */
     public static function setPermissionModel(PermissionInterface $model) {
-        self::$model = $model;
         common_ext_ExtensionsManager::singleton()->getExtensionById('generis')->setConfig(self::CONFIG_KEY, $model);
```
antoinerobin: why don't you use the service manager register here ?

jbout: Because registry ONLY works with services and some permission model implementations might not extend configurableService

PR #311, opened 2017-03-01
Feature/json keyvalue cache

bugalot:
**FULL PR SET IS
https://github.com/oat-sa/generis/pull/311
https://github.com/oat-sa/tao-core/pull/1212
https://github.com/oat-sa/extension-tao-outcome/pull/65
https://github.com/oat-sa/extension-tao-outcomeui/pull/92**

**This PR enables the KeyValueCache to use Redis as its storage.** Without this, we are not able to share the generis cache efficiently amongst servers, which is a show stopper for the TAO Cloud. Indeed, as funcAcls, roles, ... are depending on the cache, it seems obvious it has to be shared amongst web servers to provide large-scale authoring platforms (a.k.a. backoffice).

This is amazing how your TAO platform will be more reactive than ever, especially when entering the Item Authoring or taking a Delivery.

I'm aware that it might not work with all projects (ACT ?), but with TAO Vanilla it does pretty well! However, it's a config to change so no impact is expected. I will dig further on every project to see how feasible it is to go to KeyValueCache using Redis for customer projects.

In order to enable the KeyValueCache in redis:

1. Change your `config/generis/cache.conf.php` file to

```php
<?php
return new common_cache_KeyValueCache(array(
    'persistence' => 'cache',
    'jsonify' => true
));
```

2. Change your config/generis/persistences.conf.php file to

```php
<?php
return new common_persistence_Manager(array(
    'persistences' => array(
        'cache' => array(
            'driver' => 'phpredis',
            'host' => '127.0.0.1',
            'port' => 6379
        ),
        'default' => array(
            'driver' => 'pdo_pgsql',
            'host' => 'localhost',
            'dbname' => 'taoplatform300',
            'user' => 'taopostgres',
            'password' => '******'
        ),
        'serviceState' => array(
            'driver' => 'phpfile'
        )
    )
));
```


antoinerobin (commented):
Shouldn't we have something that automatically jsonify the cache when it is a redis cache in order to avoid mistakes ?

PR #326, opened 2017-03-27
abstract payload for queue

antoinerobin:


jbout:
dependencies with Tao need to be resolved

PR #332, opened 2017-04-06
Throw exception on write fail

jbout:


common/ext/class.Extension.php
```diff
@@ -148,11 +148,14 @@ public function hasConfig($key)
...
-     * @return boolean
+     * @throws common_exception_Error On error
      */
     public function setConfig($key, $value)
     {
-        return $this->getConfigPersistence()->set($this->getId().'/'.$key, $value);
+        $success = $this->getConfigPersistence()->set($this->getId().'/'.$key, $value);
+        if (!$success) {
+            throw new common_exception_Error('Unable to write '.$this->getId().'/'.$key);
+        }
```
jsconan: Maybe should you still return a value. I found some code that is using the return value of the `setConfig()` method.

https://github.com/search?l=&p=6&q=org%3Aoat-sa+setConfig+language%3APHP&ref=advsearch&type=Code&utf8=%E2%9C%93

jbout: Agreed, safer in case someone actually verified

PR #356, opened 2017-05-22
Feature/tao 3872 logger service

siwane:


common/oatbox/log/LoggerService.php
```diff
@@ -0,0 +1,71 @@
...
+     * If $replace is set to true, old logger is replaced
+     *
+     * @param LoggerInterface $logger
+     * @param boolean $replace
+     * @return LoggerInterface
+     */
+    public function addLogger(LoggerInterface $logger, $replace = false)
+    {
+        if (!$replace && (($currentLogger = $this->getLogger()) !== false)) {
+            $logger = new LoggerAggregator([$logger, $currentLogger]);
```
chilly86: What if the current logger is an instance of LoggerAggregator?

siwane: It works :) , LoggerAggregator is an instance of LoggerInterface, could be passed as another LoggerAggregator

PR #366, opened 2017-06-08
Feature/tao 4449/uniformize sync async tasks

chrisphpgarcia:
related tohttps://github.com/oat-sa/lib-oatbox-taskqueue/pull/31
https://github.com/oat-sa/extension-tao-mpart/pull/612


common/oatbox/task/AbstractQueue.php
```diff
@@ -0,0 +1,161 @@
...
+     */
+    protected $persistence;
+
+
+    public function __construct(array $options = array())
+    {
+        parent::__construct($options);
+        if($this->hasOption('runner')) {
+            $classRunner       = $this->getOption('runner');
+            $this->runner      = new $classRunner();
```
antoinerobin: You should check for the interface of the runner

common/oatbox/task/AbstractQueue.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        if($this->hasOption('persistence') && $this->hasOption('config')) {
+            $classPersistence = $this->getOption('persistence');
+            $configPersistence = $this->getOption('config');
+            $this->persistence = new $classPersistence($configPersistence);
```
antoinerobin: you should check if the persitence is the right interface and the config is an array

common/oatbox/task/AbstractQueue.php
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public function getPayload($currentUserId = null)
+    {
+        $class = $this->getOption('payload');
+        $payload = new $class($this->getPersistence() , $currentUserId);
+        $payload->setServiceLocator($this->getServiceLocator());
+        /**
+         * @var TaskPayLoad $payload
+         */
+        return $payload;
```
antoinerobin: Why don't you have this in the construct

common/oatbox/task/AbstractQueue.php
```diff
@@ -0,0 +1,161 @@
...
+
+namespace oat\oatbox\task;
+
+use oat\oatbox\service\ConfigurableService;
+use oat\oatbox\task\TaskInterface\TaskPayLoad;
+use oat\oatbox\task\TaskInterface\TaskPersistenceInterface;
+use oat\oatbox\task\TaskInterface\TaskQueue;
+use oat\oatbox\task\TaskInterface\TaskRunner;
+
+abstract class AbstractQueue
```
antoinerobin: Could you add some documentation (class and methods)

common/oatbox/task/AbstractTaskService.php
```diff
@@ -0,0 +1,119 @@
...
+
+namespace oat\oatbox\task;
+
+use oat\oatbox\service\ConfigurableService;
+use oat\oatbox\task\TaskInterface\TaskListIterator;
+use oat\oatbox\task\TaskInterface\TaskPersistenceInterface;
+use oat\oatbox\task\TaskInterface\TaskQueue;
+use oat\oatbox\task\TaskInterface\TaskRunner;
+
+abstract class AbstractTaskService
```
antoinerobin: Can't you implement these method in the other abstract class ?
It is dangerous to have two similar implementation of the same interface

PR #369, opened 2017-06-09
Fix/doctrine dependencies

jsconan:
Fix issue with `composer update` and Doctrine dependencies.


jsconan:
This patch is not needed anymore as the issue seems to be a temporary one, because if you:
- apply the patch, say just add `"doctrine/inflector": "~1.0.0",` in you composer
- run `composer update`
- remove the patch, say just remove the previously added line
- run again `composer update`
that will work!

PR #397, opened 2017-09-21
Feature/tao 4972/new task queue

gyszucs:
Having AWS SQS implementation properly, we had to re-design the Task Queue from the ground.
I would suggest reading the [README](https://github.com/oat-sa/generis/blob/feature/TAO-4972/new-task-queue/common/oatbox/TaskQueue/README.md) about the Task Queue first.


common/oatbox/TaskQueue/AbstractTask.php
```diff
@@ -0,0 +1,90 @@
...
+namespace oat\oatbox\TaskQueue;
+
+/**
+ * Class AbstractTask
+ *
+ * @author Gyula Szucs <gyula@taotesting.com>
+ */
+abstract class AbstractTask extends Message implements TaskInterface
+{
+    private $parameters = [];
```
hutnikau: Should it be private? Classes which will extend `AbstractTask` will be not able to work with this property.

common/oatbox/TaskQueue/MessageBroker/AbstractMessageBroker.php
```diff
@@ -0,0 +1,221 @@
...
+     * Denormalize the given message JSON.
+     *
+     * If the json is not valid, it deletes the message straight away without processing it.
+     *
+     * @param string $messageJSON
+     * @param string $receiptForDeletion An identification of the given message
+     * @param array  $logContext
+     * @return null|MessageInterface
+     */
+    protected function denormalizeMessage($messageJSON, $receiptForDeletion, array $logContext = [])
```
hutnikau: It seems more like unserialize (or decode) than denormalize

common/oatbox/TaskQueue/Message.php
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public function setCreatedAt(\DateTime $dateTime)
+    {
+        $this->setMetadata('_created_at_', $dateTime);
+
+        return $this;
+    }
+
+    /**
+     * @return \DateTime
```
hutnikau: I guess after call of `denormalizeMessage` it will be array but not `\DateTime`

hutnikau (changes requested):

gyszucs:
Closing it because more work on it is required and will be split into smaller PRs.

PR #401, opened 2017-09-28
Fix/tao 5058/remove lstat warning from config

ionutpad:
exists method in File class should check for existing of a path before checking the type


siwane:
Does the unit test still pass? Have you test it?

ionutpad:
Yes. unit tests passed.

PR #410, opened 2017-10-12
add php code serializer trait and renderer class

chilly86:


ionutpad:
This can be a helpful thing to have, but can you add unit tests ? Just to have prove that it's working well, and also we can test different scenarios.

PR #417, opened 2017-11-07
Feature/serializable service manager

bugalot:
This PR implements a serializable ServiceManager. Please see code comments for an explanation.


siwane:
See https://github.com/oat-sa/tao-core/pull/1540

PR #449, opened 2018-01-12
Fix/sql platform now expression

hutnikau:
I found that on ACT prod server all the time values in task queue tables (such as `tq_background`, `tq_task_log` etc) are stored in their local timezone. It's better to have all the time values in UTC in the database to avoid confusion.

### DIFF
--- a/common/persistence/sql/class.Platform.php
+++ b/common/persistence/sql/class.Platform.php
@@ -151,7 +151,7 @@ class common_persistence_sql_Platform {
      * @return string
      */
     public function getNowExpression(){
-        $datetime = new DateTime();
+        $datetime = new DateTime('now', new \DateTimeZone('UTC'));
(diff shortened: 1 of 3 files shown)


common/persistence/sql/class.Platform.php
```diff
@@ -151,7 +151,7 @@ public function getName(){
      * @return string
      */
     public function getNowExpression(){
-        $datetime = new DateTime();
+        $datetime = new DateTime('now', new \DateTimeZone('UTC'));
```
gyszucs: There is a TIME_ZONE global constant defined in generis.conf.php. I think, it would be nice to use that here.

jbout: times in database should always be UTC

gyszucs: okay, I see. that global can be used for other reasons.

PR #457, opened 2018-02-02
Feature/tao 5898 added viewable hiddenbox widget type

no-chris:


scripts/update/Updater.php
```diff
@@ -411,12 +411,23 @@ public function update($initialVersion) {
...
 
         $this->skip('4.4.1', '6.7.0');
+
+        if ($this->isVersion('6.7.0')) {
+            $file = __DIR__ . DIRECTORY_SEPARATOR .
+                '..'.DIRECTORY_SEPARATOR .'..'.DIRECTORY_SEPARATOR .
+                'core' . DIRECTORY_SEPARATOR .
+                'ontology' . DIRECTORY_SEPARATOR .
+                'generis.rdf';
+            $api = core_kernel_impl_ApiModelOO::singleton();
```
antoinerobin: Why don't you use `OntologyUpdater::syncModels();` it will add/remove all your changes in the rdf files

no-chris: see comment above ;)

PR #464, opened 2018-02-16
Feature/tao 5450/logger improvements

chilly86:


common/oatbox/log/logger/formatter/TaoJsonLogFormatter.php
```diff
@@ -0,0 +1,143 @@
...
+     * @inheritdoc
+     *
+     * @throws \ErrorException
+     */
+    public function format(array $record)
+    {
+        $jsonString = json_encode($this->getOutputRecord($record));
+
+        if ($jsonString === false) {
+            throw new \ErrorException('Error happened during the log format process! (json encode error)');
```
gyszucs: Maybe it can be worth thinking about whether we do need an exception here.

PR #467, opened 2018-03-01
Feature/tao 6000/adding script modes with legacy support

chilly86:


common/oatbox/extension/script/ScriptAction.php
```diff
@@ -51,180 +77,274 @@
...
-            );
-        }
 
-        // Run the userland script.
-        $report = $this->run();
+            // Initializes the trait options.
+            $report = $this->initializeTraitOptions($report);
+
+            // Run the userland script.
+            $report->add($this->run());
```
bugalot: As explained on slack, the *implementer MUST be in control of the generation of the initial report*. The main reason is that calling scripts might want to rely on the data returned via the report. With your change, the [following script](https://github.com/oat-sa/extension-tao-operations/blob/master/scripts/tools/CreateClusterBackofficeUsers.php#L171) will not work anymore because it is waiting for some data from the root report of another script.

I appreciate the try, but please go back to the initial pull request where the modifications I made make the legacy stuff able to continue working.

PR #469, opened 2018-03-05
Add reports to the extension update output

hutnikau:
Sometimes it's useful to add some notes to the updater script output. For example to not forger remove cron job, check that storage exists or something else.

With this PR it could be done like this:
```
$this->addReport(new Report(Report::TYPE_INFO, 'Do not forget to remove TerminatePausedAssessment cron job'));
```

Updater output:
image


hutnikau:
Removed redundant report level. Now it looks like:
image

PR #471, opened 2018-03-06
Feature/tao 6012/optional ttl mode in filecache

chilly86:


common/persistence/class.PhpFileDriver.php
```diff
@@ -78,190 +98,312 @@ class common_persistence_PhpFileDriver implements common_persistence_KvDriver, c
...
+                static::ENTRY_EXPIRATION => $this->calculateExpiresAt($ttl),
+            ];
         } else {
-            $filePath = $this->getPath($id);
-            $dirname = dirname($filePath);
-            if (!file_exists($dirname)) {
-                mkdir($dirname, self::DEFAULT_MASK, true);
+            if (null !== $ttl) {
+                throw new common_exception_NotImplemented('TTL not implemented in '.__CLASS__);
+            } elseif ($nx) {
```
jbout: Please move $nx check outside of isTtlMode()

common/persistence/class.PhpFileDriver.php
```diff
@@ -78,190 +98,312 @@ class common_persistence_PhpFileDriver implements common_persistence_KvDriver, c
...
+     * Writes the cache file.
+     *
+     * @param $id
+     * @param $value
+     *
+     * @return bool
+     *
+     * @throws \common_exception_Error
+     */
+    private function writeFile($id, $value, $preWriteCallable = [])
```
jbout: please add description of preWriteCallable to phpdoc

common/persistence/class.PhpFileDriver.php
```diff
@@ -304,12 +448,61 @@ protected function sanitizeReadableFileName($key) {
...
+    }
+
+    /**
+     * Returns the requested element from cache or FALSE if it does not exist.
+     *
+     * @param $id
+     *
+     * @return bool|mixed
+     */
+    protected function getFromCache($id)
```
jbout: Please throw an exception if the value does not exist

common/persistence/class.PhpFileDriver.php
```diff
@@ -78,190 +98,312 @@ class common_persistence_PhpFileDriver implements common_persistence_KvDriver, c
...
-    
+
+    /**
+     * Returns the raw cache record.
+     *
+     * @param $id
+     *
+     * @return mixed
+     */
+    public function getProcessedEntry($id)
```
jbout: Could this be protected?

common/persistence/class.PhpFileDriver.php
```diff
@@ -78,190 +98,312 @@ class common_persistence_PhpFileDriver implements common_persistence_KvDriver, c
...
-            // OPcache workaround
-            unset($this->cache[$id]);
-        }
+    public function del($id)
+    {
         $filePath = $this->getPath($id);
-        
+
+        // OPcache workaround
+        unset($this->cache[$id]);
```
jbout: could cause a warning if not set

test/tools/FakeGlobalFunctionTrait.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Copyright (c) 2018 (original work) Open Assessment Technologies SA;
+ *
+ *
+ */
+
+namespace oat\generis\test\tools;
+
+
+use phpmock\MockBuilder;
```
jbout: Please add dependency to require-dev of composer.json or remove

common/persistence/class.PhpFileDriver.php
```diff
@@ -78,190 +98,352 @@ class common_persistence_PhpFileDriver implements common_persistence_KvDriver, c
...
-        $this->directory = isset($params['dir']) 
+        $this->directory = isset($params['dir'])
             ? $params['dir'].($params['dir'][strlen($params['dir'])-1] === DIRECTORY_SEPARATOR ? '' : DIRECTORY_SEPARATOR)
             : FILES_PATH.'generis'.DIRECTORY_SEPARATOR.$id.DIRECTORY_SEPARATOR;
         $this->levels = isset($params['levels']) ? $params['levels'] : self::DEFAULT_LEVELS;
         $this->humanReadable = isset($params['humanReadable']) ? $params['humanReadable'] : false;
+
+        // Sets ttl mode TRUE when the passed ttl mode is true.
+        $this->setTtlMode(
+            (isset($params[static::OPTION_TTL]) && $params[static::OPTION_TTL] == true)
```
gyszucs: Why do we need an optional parameter to enable ttl? Don't we want this solution globally?

jbout: the data format changes when using TTL, so in order to be backward compatible we need to either implement it as a new solution, or make it a parameter

gyszucs: okay, I see.

PR #477, opened 2018-03-13
Tao 5862 redis full storage error processing

hutnikau:


common/persistence/interface.KvDriver.php
```diff
@@ -34,12 +33,14 @@
     /**
      * Stores a value, implementing time to live and nx is optional
      * Should throw an exception if an option is not supported
-     * 
+     *
      * @param string $id
      * @param string $value
      * @param string $ttl time to live in seconds
      * @param bool $nx Only set the key if it does not already exist
      * @return bool
```
jbout: What does the bool return value indicate if errors are now handled via exceptions?

scripts/update/Updater.php
```diff
@@ -297,6 +297,6 @@ public function update($initialVersion) {
             $this->setVersion('6.9.0');
         }
 
-        $this->skip('6.9.0', '6.15.0');
+        $this->skip('6.9.0', '6.15.1');
```
jbout: Changing the interface can cause breaking changes, some that I spotted:
* tao_models_classes_service_StateStorage::set()
* ServiceManager::register()
* ExtendedState::save()
* StorageManager::persistCacheEntry()

PR #481, opened 2018-03-19
Fix/undefined index test error

ionutpad:


core/user/AuthAdapter.php
```diff
@@ -121,10 +121,12 @@ public function authenticate() {
 	        throw new core_kernel_users_InvalidLoginException('Invalid password for user "'.$this->username.'"');
 	    }
 
-        $userFactory = ServiceManager::getServiceManager()->get($this->options['user_factory']) ;
+	    if (isset($this->options['user_factory'])){
```
siwane: you can use this->addOption

PR #499, opened 2018-04-19
Directory::rename method

bugalot:


common/oatbox/filesystem/Directory.php
```diff
@@ -144,4 +144,9 @@ protected function getFullPath($path)
     {
         return $this->getPrefix() . '/' . $this->sanitizePath($path);
     }
+
+    public function rename($path)
```
siwane: Could u add PHP doc please.
Also, can you catch ` FileExistsException` and `FileNotFoundException` to render a tao exception

PR #522, opened 2018-08-01
Fix/monolog version

jbout:


composer.json
```diff
@@ -63,7 +63,7 @@
 		"php" : ">=5.3.2",
 		"psr/log": "~1.0",
 		"oat-sa/lib-generis-search": "2.0.1",
-		"monolog/monolog": "^1.24.0",
+		"monolog/monolog": "^1.23",
 		"fluent/logger": "^0.3.8"
```
chilly86: the interface is different between 1.x and 0.3.8, but the only method what we use (post) has the same interface, so it shouldn't break if we use the most recent one.

PR #534, opened 2018-09-26
Refactor/password constraint service

jbout:


SashaTAO:
As we discussed with @jbout  it can not be merged in current state.

PR #535, opened 2018-10-01
Fix/ttl mode on windows

zagovorichev:
The issue on the Windows, when `$entry` is null and `$entry[static::ENTRY_EXPIRATION]` is not set.


common/persistence/class.PhpFileDriver.php
```diff
@@ -202,8 +202,11 @@ private function writeFile($id, $value, $preWriteValueProcessor = null)
     public function get($id)
     {
         $entry = $this->readFile($id);
-        if ($entry != false && $this->isTtlMode()) {
-            $entry = (is_null($entry[static::ENTRY_EXPIRATION]) || $entry[static::ENTRY_EXPIRATION] > $this->getTime())
+
+        if ($entry !== false && $this->isTtlMode()) {
```
zagovorichev:
```suggestion
        if ($entry !== false && $this->isTtlMode()) {
```
@jbout Please, confirm that I can change this behavior: before it worked like this: when file has `return null` - driver return `null ` as a result, but now in such situation it returns 'false'

zagovorichev:
@jbout I've added unit tests, check this, please

test/integration/common/PhpFileDriverTest.php
```diff
@@ -0,0 +1,84 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018  (original work) Open Assessment Technologies SA;
+ *
+ * @author Alexander Zagovorichev <zagovorichev@1pt.com>
+ */
+
+class PhpFileDriverTest extends PHPUnit_Framework_TestCase
```
jbout: please add a namespace to prevent collisions, and this test should work even if tao is not installed, could you consider it a unit test?

jbout (changes requested):

test/integration/common/PhpFileDriverTest.php
zagovorichev: yes, I can consider it is a unit test and this test will work even if tao is not installed. You don't like that I read real files?

PR #537, opened 2018-10-04
Fix/php7.2 compatibility

boajer:
As per changes in php 7.2
`An E_WARNING will now be emitted when attempting to count() non-countable types (this includes the sizeof() alias function).`
So in order to maximise compatibility with  following style https://github.com/oat-sa/generis/blob/b155a1c30ea08096e753d3d261dd459c5ad7c554/common/oatbox/extension/script/ScriptAction.php#L179


jbout:
This should be fixed in the usage of the report instead.

PR #538, opened 2018-10-05
Tao 6860/clean out all the file statements linked to non existent files

boajer:
https://oat-sa.atlassian.net/browse/TAO-6860

Cleans additionally to specified in story

Files that no more stored as per oat-sa/extension-tao-item#327


common/report/class.Report.php
```diff
@@ -337,4 +337,13 @@ public function toArray()
             }, $this->elements)
         ]);
     }
+
+    /**
+     * @return int
+     */
+    public function count()
```
zagovorichev: do you really need this?

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+        $this->getLogger()->info($message);
+        $this->remove($resource);
+    }
+
+    /**
+     * @param core_kernel_classes_Resource $resource
+     * @param FileSystemHandler $file
+     */
+    protected function manageOrpane(core_kernel_classes_Resource $resource, FileSystemHandler $file)
```
zagovorichev: `manageOrphan`?

zagovorichev (changes requested):

zagovorichev (approved):

siwane:
@martijn-tao as you have done same kind of script, could u try to harmonize both?

martijn-tao:
> @martijn-tao as you have done same kind of script, could u try to harmonize both?

For my script we've introduced a ResourceCollection which could be used here as well. 
It might also be a good idea to split some functionality to make the script slightly less complex.

More details in my PR: https://github.com/oat-sa/generis/pull/543

boajer:
@martijn-tao I really like your ResourceCollection. Looks like https://github.com/oat-sa/generis/pull/543 should be stabilised and merged first, before redoing my script.

martijn-tao:
> @martijn-tao I really like your ResourceCollection. Looks like #543 should be stabilised and merged first, before redoing my script.

@boajer Thanks! Sounds like a good idea. That PR should almost be done _(I believe)_. The current ResourceCollection shouldn't change anymore, so you could already pull that, and work on a refactor using it in the meantime.

martijn-tao:
@boajer My work got merged, so the ResourceCollection is now available 🎉

bartlomiejmarszal:
I fix only part of this script that was responsible for using different service to collect statements that were not available in filesystem

ekkinox (approved):

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -168,10 +168,8 @@ private function getRedundantFiles()
      */
     private function markForRemoval(core_kernel_classes_Resource $resource)
     {
-        if ($this->wetRun) {
```
boajer: Why you've removed wet\dry run support?

tikhanovichA (approved):

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -0,0 +1,239 @@
...
+    private $wetRun = false;
+    private $verbose = false;
+    private $removedCount = 0;
+    private $redundantCount = 0;
+    private $affectedCount = 0;
+    private $errorsCount = 0;
+    private $report;
+    private $markedTobeRemoved =[];
+
+    protected function provideOptions()
```
martijn-tao: Some DocBlocks are missing

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -0,0 +1,239 @@
...
+    {
+        return $this->getServiceLocator()
+            ->get(\common_persistence_Manager::SERVICE_ID)
+            ->getPersistenceById('default');
+    }
+
+    private function getRedundantFiles()
+    {
+        return [
+            'qti.xml' //special case, see linked story ( has been stored at RDS, but never referenced via resource ).
```
martijn-tao: If this method had a docblock, that would be a perfect place for this comment ;)

martijn-tao (changes requested):

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -0,0 +1,239 @@
...
+        return true;
+    }
+
+    /**
+     * @param core_kernel_classes_Resource $resource
+     * @return bool
+     */
+    private function isOrphan(core_kernel_classes_Resource $resource)
+    {
+        $sql = 'SELECT subject FROM statements s WHERE s.object=?';
```
siarhei-baradzin: `SELECT 1 ... LIMIT 1` could improve performance if it matters here

siarhei-baradzin (approved):

scripts/tools/CleanUpOrphanFiles.php
bartlomiejmarszal: We are going to remove this script anyway. Is there a point to add docblocks now?

scripts/tools/CleanUpOrphanFiles.php
bartlomiejmarszal: This is a one-time action. I don't think that this is important here.

scripts/tools/CleanUpOrphanFiles.php
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    private function getPersistence()
+    {
+        return $this->getServiceLocator()
+            ->get(\common_persistence_Manager::SERVICE_ID)
+            ->getPersistenceById('default');
+    }
+
+    private function getRedundantFiles()
```
siarhei-baradzin: Pls pay attention this method isn't used, ignore my notice if it's ok (I'm not aware of the whole task)

martijn-tao (approved):

martijn-tao:
🙈

PR #539, opened 2018-10-08
Feature/use url fileserializer

jbout:


hutnikau:
`UrlFileSerializer` stores path as literal, but the most properties (e.g. `http://www.tao.lu/Ontologies/TAOItem.rdf#ItemContent`) expects this value to be resource with `http://www.tao.lu/Ontologies/generis.rdf#File` type. Is it acceptable?

siwane:
The `http://www.tao.lu/Ontologies/generis.rdf#File` property is normally managed by the resource file serializer. If nothing has been forgotten during refactoring, it should be transparent

PR #540, opened 2018-10-11
PDO replaced with DBAL

alexeyvy:
According to https://oat-sa.atlassian.net/browse/TAO-7005


alexeyvy:
Closed, as the code affected line endings

PR #542, opened 2018-10-12
PDO replaced with DBAL

alexeyvy:
Attempt #2, after `git config --global core.autocrlf false`


scripts/update/Updater.php
```diff
@@ -331,5 +331,28 @@ public function update($initialVersion) {
         }
 
         $this->skip('7.2.0', '7.9.8');
+
+        if ($this->isVersion('7.9.8')) {
+            /** @var common_ext_ExtensionsManager $extensionManager */
+            $extensionManager = $this->getServiceManager()->get(common_ext_ExtensionsManager::SERVICE_ID);
+            $persistenceManager = $extensionManager->getExtensionById('generis')->getConfig('persistences');
```
jbout: Should be retrieve in the same way as the extension manager

core/kernel/classes/class.DbWrapper.php
```diff
@@ -429,14 +429,15 @@ public function createIndex($indexName, $tableName, $columns){
     }
 
     /**
-     * Rebuild the indexes of a given table. This method throws PDOExceptions in
+     * Rebuild the indexes of a given table. This method throws DBALException in
      * of error.
```
jbout: in "case" of error

scripts/update/Updater.php
```diff
@@ -350,5 +350,27 @@ public function update($initialVersion) {
...
+
+        if ($this->isVersion('8.1.0')) {
+            /** @var \common_persistence_Manager $persistenceManager */
+            $persistenceManager = $this->getServiceManager()->get(\common_persistence_Manager::SERVICE_ID);
+
+            $persistenceManagerConfig = $persistenceManager->getOption('persistences');
+
+            $pdoDriverPrefix = 'pdo_';
+            foreach ($persistenceManagerConfig as $persistenceId => $persistenceParams) {
+                if (substr($persistenceParams['driver'], 0, strlen($pdoDriverPrefix)) === $pdoDriverPrefix) {
```
hutnikau: `strpos($persistenceParams['driver'], $pdoDriverPrefix) === 0` a bit shorter and readable :)

core/kernel/persistence/smoothsql/class.Class.php
```diff
@@ -471,7 +472,7 @@ public function deleteInstances( core_kernel_classes_Class $resource, $resources
         		// as deleted.
         		$this->getPersistence()->exec($query);	
         		$returnValue = true;
-            } catch (PDOException $e) {
+            } catch (DBALException $e) {
```
hutnikau: While PDO driver class still exists in Tao catching only `DBALException` may break some functionality. I would suggest two options:
1. Catch both of them
2. Log a warning message (which says that deprecated driver is configured) for any initialization of PDO driver in order to warn that it may lead to some inconsistent behavior.

jbout: no longer exists in TAO

PR #543, opened 2018-10-12
Script to migrate files from the ResourceFileSerializer to the UrlFileSerializer

martijn-tao:
Usage: `php index.php '\oat\generis\scripts\tools\FileSerializerMigration\Migrate' [--wet-run]`

- Use the `--wet-run` parameter to do a "wet" run.
- ~Use the `--limit` parameter to define how many items should be processed in one batch (default is set to 200)~


martijn-tao:
Possible todo would be implementing a chunker to process migration in chunks. Will wait for feedback on wether or not that's needed

martijn-tao:
Quick update. I did a little benchmarking, and currently migrating 700 files will result in the following stats:
**Dry run:** ~12 seconds
**Wet run:** ~30 seconds

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,195 @@
...
+     * @throws \oat\generis\model\fileReference\FileSerializerException
+     * @throws \oat\search\base\exception\SearchGateWayExeption
+     * @throws common_exception_Error
+     */
+    private function migrate()
+    {
+        $oldResourcesData = $this->migrationHelper->getOldResourcesData();
+        $oldResourceCount = $this->migrationHelper->migrationInformation['old_resource_count'];
+
+        $chuncks = array_chunk($oldResourcesData, self::RESOURCE_LIMIT);
```
jbout: the idea of chunking is not to load all resources into memory, not sure there is a benefit of chunking the data in this case as you will only iterate over the chunk in the helper

test/unit/helpers/FileSerializerMigrationHelperTest.php
```diff
@@ -0,0 +1,130 @@
...
+    public function setUp()
+    {
+        $this->fileMigrationHelper = new FileSerializerMigrationHelper(true);
+        $this->urlFileSerializer = new UrlFileSerializer();
+        $this->resourceFileSerializer = new ResourceFileSerializer();
+
+        $currentFileSerializer = ServiceManager::getServiceManager()->get(FileReferenceSerializer::SERVICE_ID);
+        if ($currentFileSerializer instanceof UrlFileSerializer) {
+            $this->fileSerializerSwitched = true;
+            ServiceManager::getServiceManager()->register(FileReferenceSerializer::SERVICE_ID, new ResourceFileSerializer());
```
jbout: unit tests should never change the server config

test/unit/helpers/FileSerializerMigrationHelperTest.php
```diff
@@ -0,0 +1,130 @@
...
+
+use core_kernel_classes_Class;
+use core_kernel_classes_Resource;
+use oat\generis\Helper\FileSerializerMigrationHelper;
+use oat\generis\model\fileReference\FileReferenceSerializer;
+use oat\generis\model\fileReference\ResourceFileSerializer;
+use oat\generis\model\fileReference\UrlFileSerializer;
+use oat\generis\test\GenerisPhpUnitTestRunner;
+use oat\oatbox\service\ServiceManager;
+use oat\taoDevTools\helper\DataGenerator;
```
jbout: We cannot introduce a dependency to taoDevTools in tao

test/unit/helpers/FileSerializerMigrationHelperTest.php
```diff
@@ -0,0 +1,130 @@
...
+
+    }
+
+    /**
+     * Generate 10 file resources used for testing
+     */
+    private function generateFileResources()
+    {
+        if ($this->testItems !== null) {
+            $this->tearDown();
```
jbout: tearDown should not be called manually

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,195 @@
...
+                'prefix' => 'w',
+                'longPrefix' => 'wet-run',
+                'description' => 'Add this flag to migrate the references, as opposed to just listing them (dry/wet run)'
+            ],
+            'resourceLimit' => [
+                'prefix' => 'l',
+                'longPrefix' => 'limit',
+                'cast' => 'integer',
+                'defaultValue' => self::RESOURCE_LIMIT,
+                'description' => 'How many items should be processed each run?'
```
jbout: This could mean that we will only run the script on the first X resources

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,195 @@
...
+     * Get the file serializer migration helper
+     *
+     * @return FileSerializerMigrationHelper
+     * @throws InvalidServiceManagerException
+     */
+    private function getMigrationHelper()
+    {
+        if ($this->migrationHelper === null) {
+            $this->migrationHelper = new FileSerializerMigrationHelper($this->isWetRun());
+            $this->migrationHelper->setServiceLocator($this->getServiceLocator());
```
gyszucs: Probably you don't need both services. The service manager should be enough.

core/fileReference/ResourceFileIterator.php
```diff
@@ -0,0 +1,283 @@
...
+
+    /**
+     * Get the file resources
+     *
+     * @return mixed[]
+     */
+    private function getFileResources()
+    {
+        $sql = "
+            SELECT `subject`, `id` FROM ((
```
gyszucs: Don't use backtick please. Postgresql doesn't like it.

core/fileReference/ResourceFileIterator.php
```diff
@@ -0,0 +1,283 @@
...
+                SELECT DISTINCT `id`, `subject` FROM `statements`
+                    WHERE  `predicate` = '" . OntologyRdf::RDF_TYPE . "'
+                    AND (`object` = '" . GenerisRdf::CLASS_GENERIS_FILE . "')
+                    AND `id` > " . $this->lastId . "
+                    AND `id` NOT IN('" . implode("', '", array_keys($this->corruptFileIds)) . "')
```
gyszucs: Postgresql also doesn't like this: `SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: ""` It happens at the first run because `$this->corruptFileIds` is empty. Try to use something like this: https://github.com/oat-sa/tao-core/blob/master/models/classes/taskQueue/TaskLog/Broker/RdsTaskLogBroker.php#L353 but using `Connection::PARAM_INT_ARRAY` instead of `Connection::PARAM_STR_ARRAY`

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,195 @@
...
+
+        $migrationHelper->migrateFiles(
+            $this->hasOption('resourceLimit') ? $this->getOption('resourceLimit') : self::RESOURCE_LIMIT
+        );
+
+        $this->addMigrationReport();
+
+        if ($migrationHelper->updateFileSerializer()) {
+            $this->report->add(Report::createSuccess(
+                'Successfully updated FileReferenceSerializer service to use UrlFileSerializer service'
```
gyszucs: This one is displayed even in dry-run. That can be confusing.

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,195 @@
...
+     * Add the reports related to the migration
+     *
+     * @throws common_exception_Error
+     * @throws InvalidServiceManagerException
+     */
+    private function addMigrationReport()
+    {
+        $migrationHelper = $this->getMigrationHelper();
+        $this->report->add(Report::createSuccess(sprintf(
+            'Successfully migrated %s old references to %s new file serializer references.',
```
gyszucs: How about changing this sentence for dry-run? It's also a bit confusing.

gyszucs:
After some testing, the PR works for me. @jbout How about you?

scripts/update/Updater.php
```diff
@@ -28,6 +28,7 @@
 use oat\generis\model\data\ModelManager;
 use oat\generis\model\fileReference\FileReferenceSerializer;
 use oat\generis\model\fileReference\ResourceFileSerializer;
+use oat\generis\model\fileReference\UrlFileSerializer;
```
jbout: probably not required anymore?

test/unit/helpers/FileSerializerMigrationHelperTest.php
```diff
@@ -0,0 +1,220 @@
...
+     * @var FileSystemService
+     */
+    private $fileSystemService;
+
+    /**
+     * Initialize test
+     */
+    public function setUp()
+    {
+        common_Config::load();
```
jbout: Why would you need to load the config of the current install in a unit test?

core/fileReference/ResourceFileIterator.php
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * @var int
+     */
+    private $lastId = 0;
+
+    /**
+     * @var int[]
+     */
+    private $corruptFileIds = [];
```
jbout: What is the purpose of corruptFileIds?

scripts/tools/FileSerializerMigration.php
```diff
@@ -0,0 +1,219 @@
...
+     * @throws common_exception_Error
+     */
+    protected function run()
+    {
+        $this->report = Report::createInfo('Starting file serializer migration.');
+
+        $migrationHelper = $this->getMigrationHelper();
+
+        $migrationHelper->migrateFiles(
+            $this->hasOption('resourceLimit') ? $this->getOption('resourceLimit') : self::RESOURCE_LIMIT
```
jbout: this is called resourceLimit here, but gets ibnterpreted as cache size in the Helper, could be confussing

helpers/FileSerializerMigrationHelper.php
```diff
@@ -0,0 +1,253 @@
...
+        ++$this->migrationInformation['migrated_count'];
+    }
+
+    /**
+     * Update the FileReferenceSerializer service to use the UrlFileSerializer.
+     *
+     * @return bool
+     * @throws common_Exception
+     */
+    public function updateFileSerializer()
```
jbout: Does this need to be in the helper?

scripts/tools/FileSerializerMigration/FileSerializerMigrationAction.php
```diff
@@ -189,6 +180,42 @@ private function addMigrationReport()
...
+    {
+        $updated = false;
+        if ($this->fileSerializerNeedsUpdate()) {
+            if ($this->isWetRun) {
+                $this->getServiceManager()->register(FileReferenceSerializer::SERVICE_ID, new UrlFileSerializer());
+            }
+            $updated = true;
+        }
+
+        return $updated;
```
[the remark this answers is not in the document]
martijn-tao: True, but I believe it's considered bad practice to have multiple return statements in a method. That's why I chose this option instead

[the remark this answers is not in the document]
Babacooll: In my mind (Code complete states it also) multiple return statements are acceptable if it enhance readability of the code. Here is a good example for me, but I get your point.

martijn-tao:
Current performance on local machine:
DRY RUN - 1500 file resources -> ~3 sec
WET RUN - 1500 file resources -> ~11 sec

scripts/tools/FileSerializerMigration/Migrate.php
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Check if the file serializer service needs to be updated
+     *
+     * @return bool
+     * @throws InvalidServiceManagerException
+     */
+    private function fileSerializerNeedsUpdate()
+    {
+        $needsUpdate = true;
+        $currentFileReferenceSerializer = $this->getServiceManager()->get(FileReferenceSerializer::SERVICE_ID);
```
siwane: Use serviceLocator please

scripts/tools/FileSerializerMigration/Migrate.php
```diff
@@ -0,0 +1,246 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017  (original work) Open Assessment Technologies SA;
```
siwane: Date is wrong

core/resource/ResourceCollection.php
```diff
@@ -0,0 +1,284 @@
...
+    {
+        return $this->endOfClass;
+    }
+
+    /**
+     * Adds a filter for a class
+     */
+    private function addClassFilter()
+    {
+        if (is_string($this->class)) {
```
siwane: u can directly use `getClass`, it can be constrcuted with uri or class object

core/resource/ResourceCollection.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * Check if we reached the limit of the amount of items we should process.
+     *
+     * @return bool
+     */
+    private function isLimitReached()
+    {
+        return $this->limit !== null && (!$this->count()) < $this->limit;
```
siwane: why `!`?

core/resource/ResourceCollection.php
```diff
@@ -0,0 +1,284 @@
...
+        return $this->endOfClass;
+    }
+
+    /**
+     * Adds a filter for a class
+     */
+    private function addClassFilter()
+    {
+        if (is_string($this->class)) {
+            $this->getClass($this->class);
```
siwane: does it have to be `$this->class = $this->getClass($this->class);`?

scripts/tools/FileSerializerMigration/MigrationHelper.php
```diff
@@ -0,0 +1,234 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
siwane: Wrong date

scripts/tools/FileSerializerMigration/MigrationHelper.php
```diff
@@ -0,0 +1,234 @@
...
+
+        return $this->resourceFileSerializer;
+    }
+
+    /**
+     * Set the Service Locator for this class
+     *
+     * @param ServiceLocatorInterface $serviceLocator
+     */
+    public function setServiceLocator(ServiceLocatorInterface $serviceLocator)
```
siwane: Could u used ServiceLocatorAwareTrait please?

test/unit/helpers/FileSerializerMigrationHelperTest.php
siwane: up

martijn-tao:
> Need to add ServiceLocatorAwareInterface ;)

Oops, fixed

PR #545, opened 2018-10-23
isFileInsideDirectory helper for preventing path traversal attempts

alexeyvy:


SergiiTao:
no unittests

helpers/class.File.php
```diff
@@ -95,6 +95,20 @@ static public function getRelPath($from, $to)
...
+     *
+     * @param $filename
+     * @param $directory
+     * @return bool
+     */
+    static public function isFileInsideDirectory($filename, $directory)
+    {
+        $canonicalDirectory = realpath($directory);
+        $canonicalFilename = realpath($canonicalDirectory . DIRECTORY_SEPARATOR . $filename);
+        return 0 === strpos($canonicalFilename, $canonicalDirectory);
```
SergiiTao: realpath returns FALSE in case of error. it could be a problem for strpos

PR #562, opened 2018-12-13
TAO-7277 - Add service that handles iframe protection

martijn-tao:
Related to https://github.com/oat-sa/tao-core/pull/1912


siwane:
You do not have the config at install

martijn-tao:
All logic has been moved to oat-sa/tao-core

PR #573, opened 2019-01-21
Fix/tao 6636/failover of db

zagovorichev:
1. switch to the dbal:
in the *config/generis/persistences.conf.php* change driver to the `'driver' => 'dbal_pdo_pgsql',`
2. I use xDebut to test it:
- stop postgesql
- set breakpoint https://github.com/oat-sa/generis/pull/573/files#diff-1de4d1e7ac1606acd905ee35015dc610R73
- catch an exception
- run postgresql

### DIFF
--- a/common/persistence/sql/dbal/class.Driver.php
+++ b/common/persistence/sql/dbal/class.Driver.php
@@ -64,7 +64,10 @@ class common_persistence_sql_dbal_Driver implements common_persistence_sql_Drive
 
         while (true) {
             try {
+                /** @var \Doctrine\DBAL\Connection connection */
                 $this->connection = $this->getConnection($connectionParams, $config);
+                // to generate DBALException if no connection
+                $this->connection->ping();
                 break;
             } catch (\Doctrine\DBAL\DBALException $e) {
                 $this->connection = null;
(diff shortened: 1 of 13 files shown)


common/persistence/sql/dbal/class.Driver.php
```diff
@@ -52,12 +58,60 @@ function connect($id, array $params)
         $config = new \Doctrine\DBAL\Configuration();
 //          $logger = new Doctrine\DBAL\Logging\EchoSQLLogger();
```
martijn-tao: Do we have a good reason to leave this commented code in?

martijn-tao:
@zagovorichev What would be the best way for me to test the functionality on my local machine? Maybe you could provide a small guide

common/persistence/sql/dbal/class.Driver.php
zagovorichev: someone left it here, I guess that it can simplify debugging

martijn-tao: That's what I thought, just wanted to check ;)

zagovorichev:
@martijn-tao I've added guide to the first commentary

common/persistence/sql/dbal/class.Driver.php
```diff
@@ -34,14 +34,20 @@ class common_persistence_sql_dbal_Driver implements common_persistence_sql_Drive
...
      * Connect to Dbal
      * 
      * @param string $id
      * @param array $params
-     * @return \Doctrine\DBAL\Connection;
+     * @return common_persistence_Persistence|common_persistence_SqlPersistence
+     * @throws \Doctrine\DBAL\DBALException
      */
-    function connect($id, array $params)
+    public function connect($id, array $params)
```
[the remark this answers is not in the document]
martijn-tao: @zagovorichev This could probably be fixed by moving the while loop into a new function (eg. `testConnection()`)

PR #576, opened 2019-01-29
Local extended configuration

zagovorichev:
Guide how to test:

Create *config/local.conf.json* file, with content:
```
{
  "generis": {
    "persistences": {
      "persistences": {
        "default": {
          "driver": "pdo_mysql"
        }
      }
    }
  }
}
```

that means that `driver` value will be replaced for the config `generis/persistences.conf.php`


siwane (commented):
I globally approved. However, I have couple of question: 
1. if a config is overridden by an updater, do u think it can be good idea to warn devops if the target of "extendedConf.json" has been changed?
2. Also if we want to change a ConfigurableService itself, is it possible? 
3. Last but not least, can we add a non declared config? (e.g. awsClient.conf.php)

zagovorichev:
@siwane Sorry, I've lost this PR
1. Not sure, I guess that they have to say this. (for example, they manage cron by their own, we don't see all crons' for each user in the system on the taoUpdate).
2. Everything possible, I've changed the driver, not a service.
3. Yes, it will be added, I guess it was in the AC.

PR #584, opened 2019-02-12
Memory Leak

ihar-aleksiaichuk:
memory limit
reading file

### DIFF
--- a/common/log/class.SingleFileAppender.php
+++ b/common/log/class.SingleFileAppender.php
@@ -136,11 +136,17 @@ class common_log_SingleFileAppender extends common_log_BaseAppender
     {
         if ($this->maxFileSize > 0 && file_exists($this->filename) && filesize($this->filename) >= $this->maxFileSize) {
             // need to reduce the file size
-            $file = file($this->filename);
-            $file = array_splice($file, ceil(count($file) * $this->reduceRatio));
+            $reduceLen= ($this->maxFileSize< (int)ini_get('memory_limit')*1024?$this->maxFileSize:(int)ini_get('memory_limit')*1024)*$this->reduceRatio;
+            $size = filesize($this->filename)-$reduceLen;
+            $log =file_get_contents($this->filename, false, null,  $size, $reduceLen );
+
+            $log= explode(PHP_EOL, $log);
+            //delete first element (can be broken)
+            array_shift($log);
+
             $this->filehandle = @fopen($this->filename, 'w');
-            foreach ($file as $line) {
-                @fwrite($this->filehandle, $line);
+            foreach ($log as $line) {
+                @fwrite($this->filehandle, $line.PHP_EOL);
             }
         } else {
             $this->filehandle = @fopen($this->filename, 'a');


common/log/class.SingleFileAppender.php
```diff
@@ -136,11 +136,17 @@ protected function initFile()
     {
         if ($this->maxFileSize > 0 && file_exists($this->filename) && filesize($this->filename) >= $this->maxFileSize) {
             // need to reduce the file size
-            $file = file($this->filename);
-            $file = array_splice($file, ceil(count($file) * $this->reduceRatio));
+            $reduceLen= ($this->maxFileSize< (int)ini_get('memory_limit')*1024?$this->maxFileSize:(int)ini_get('memory_limit')*1024)*$this->reduceRatio;
```
martijn-tao: For readability it might be better to change this from the shorthand if else to the full controll structure:

```
if ($this->maxFileSize < (int)ini_get('memory_limit') * 1024) {
    $reduceLength = ($this->maxFileSize) * $this->reduceRatio;
} else {
    $reduceLength = ((int)ini_get('memory_limit') * 1024) * $this->reduceRatio;
}
```

common/log/class.SingleFileAppender.php
```diff
@@ -136,11 +136,17 @@ protected function initFile()
...
+            $size = filesize($this->filename)-$reduceLen;
+            $log =file_get_contents($this->filename, false, null,  $size, $reduceLen );
```
martijn-tao: Try to be consistent with spacing. For example always surround `=` with spaces (eg. `$log = file_get_contents(...`)

common/log/class.SingleFileAppender.php
```diff
@@ -136,11 +136,17 @@ protected function initFile()
...
+
+            $log= explode(PHP_EOL, $log);
+            //delete first element (can be broken)
+            array_shift($log);
+
             $this->filehandle = @fopen($this->filename, 'w');
-            foreach ($file as $line) {
-                @fwrite($this->filehandle, $line);
+            foreach ($log as $line) {
+                @fwrite($this->filehandle, $line.PHP_EOL);
```
martijn-tao: As per the PSR-2 standard, please surround the concatenation sign (`.`) with spaces.

`$line . PHP_EOL`

common/log/class.SingleFileAppender.php
```diff
@@ -136,11 +136,17 @@ protected function initFile()
     {
         if ($this->maxFileSize > 0 && file_exists($this->filename) && filesize($this->filename) >= $this->maxFileSize) {
             // need to reduce the file size
-            $file = file($this->filename);
-            $file = array_splice($file, ceil(count($file) * $this->reduceRatio));
+            $reduceLen= ($this->maxFileSize< (int)ini_get('memory_limit')*1024?$this->maxFileSize:(int)ini_get('memory_limit')*1024)*$this->reduceRatio;
+            $size = filesize($this->filename)-$reduceLen;
```
martijn-tao: Please surround mathematical operators (`+`, `-`, `*`, `/`, `%`) with spaces

common/log/class.SingleFileAppender.php
martijn-tao: Also please omit the whitespace after the last argument (in this case `$reduceLen`)

martijn-tao:
@ihar-aleksiaichuk Could you also add a comment on how to test the added functionality?

common/log/class.SingleFileAppender.php
alexeyvy: A variable for `(int)ini_get('memory_limit') * 1024` wouldn't also be out of place here

alexeyvy:
You should probably cover this case with a unit test

PR #586, opened 2019-02-13
fixed sqlite tests

ihar-aleksiaichuk:
changed version


test/unit/common/persistence/sql/PlatformTest.php
```diff
@@ -23,7 +23,14 @@
 
 class PlatformTest extends TestCase
 {
-
+    protected function setUp()
+    {
+        if (!extension_loaded('pdo_sqlite')) {
```
alexeyvy: You can just use oat\generis\test::getSqlMock() inside test cases below. It does the same (checking the pdo_sqlite is loaded and creating connection. With this approach you don't ever need to have this setUp().

PR #598, opened 2019-03-22
Phpdoc fix

siarhei-baradzin:
Calllable -> callable in phpdoc, makes possible for PhpStorm to find usages of passed method

Actually only 1 letter changed and line endings from CRLF to LF

### DIFF
--- a/common/oatbox/event/EventManager.php
+++ b/common/oatbox/event/EventManager.php
@@ -60,7 +60,7 @@ class EventManager extends ConfigurableService
      * Attach a Listener to one or multiple events
      * 
      * @param mixed $event either an Event object or a string
-     * @param Callable $callback
+     * @param callable $callback
      */
     public function attach($event, $callback) {
         $events = is_array($event) ? $event : array($event);


common/oatbox/event/EventManager.php
```diff
@@ -1,131 +1,131 @@
...
+        }
+    }
+    
+    /**
+     * Attach a Listener to one or multiple events
+     * 
+     * @param mixed $event either an Event object or a string
+     * @param callable $callback
+     */
+    public function attach($event, $callback) {
```
julien-sebire: Why not also add callable as a typehint?

siarhei-baradzin: It sounds reasonable, but:
1. I'm not sure how it will work for `[OurService::SERVICE_ID, 'method']` which is actually not a callable in php sense, but is handled (by looking for service in ServiceManager) in `trigger()` method
2. originally I didn't want to change behavior (potentially it could cause type errors in many dependent extensions), it need to be tested, we should increase version of generis..

PR #599, opened 2019-03-22
add resource and property factories

peetya:


test/integration/core/kernel/Factory/ResourceFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+            'debug'
+        ));
+    }
+
+    /**
+     * @expectedException LogicException
+     */
+    public function testCreateMethodThrowsLogicExceptionWhenClassNotExists()
+    {
+        $this->resourceFactory->create('test', 'test');
```
Skyld: Please assert on the exception message as well

test/integration/core/kernel/Factory/ResourceFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->resourceFactory = new ResourceFactory();
+    }
+
+    /**
+     * @dataProvider fqcnDataProvider
+     * @param string $fqcn
+     */
+    public function testCreateMethodReturnsTheCorrectInstance($fqcn)
+    {
+        $this->assertInstanceOf(core_kernel_classes_Resource::class, $this->resourceFactory->create(
```
Skyld: Please assert on the uri as well

test/integration/core/kernel/Factory/ResourceFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @dataProvider fqcnDataProvider
+     * @param string $fqcn
+     */
+    public function testCreateMethodWithDebugReturnsTheCorrectInstance($fqcn)
+    {
+        $this->assertInstanceOf(core_kernel_classes_Resource::class, $this->resourceFactory->create(
+            $fqcn,
+            'test',
```
Skyld: Please assert on debug if possible, otherwise it's equal to the previous test case

test/integration/core/kernel/Factory/ResourceFactoryTest.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function testCreateMethodThrowsLogicExceptionWhenNotIsNotInstanceOfResource()
+    {
+        $this->resourceFactory->create(ResourceFactory::class, 'test');
+    }
+
+    public function fqcnDataProvider()
+    {
+        return [
+            [core_kernel_classes_Resource::class],
```
Skyld: You can list all the available resource classes here

test/integration/core/kernel/Factory/ResourceFactoryTest.php
Skyld: Also, since you are using different class instances in the dataprovider (which os a good approach), you should assert on the actual instance, not on the parent. You expect the child classes, not the generic resource class every time

peetya: True, forgot to refact this part :)

julien-sebire:
This looks good to me, but I don't think I have the write access needed to approve, so we have to find somebody who has...

PR #601, opened 2019-03-28
Fix/tao 7810/phpdocfix

siarhei-baradzin:
ComplexSearchService::searchType phpdoc fixed

No changes in behavior, only to fix IDE warnings, so I didn't change the version of the extension


ekkinox:
dont you need to bump up generis version in manifest + updater ?

ekkinox (changes requested):
dont you need to bump up generis version in manifest + updater ?

tikhanovichA:
> dont you need to bump up generis version in manifest + updater ?

Agree with Jonathan that we need to bump version

tikhanovichA (approved):

ekkinox (approved):

PR #603, opened 2019-04-09
Feature/poormans psr11

jbout:


common/oatbox/service/ServiceManager.php
```diff
@@ -63,25 +68,69 @@ public function __construct($configService)
...
+        if ($service === false) {
+            $service = $this->tryAutowiring($serviceId, $serviceKey);
         }
-        if (!isset($this->services[$serviceKey])) {
-            $service = $this->getConfig()->get($serviceKey);
-            if ($service === false) {
-                throw new ServiceNotFoundException($serviceKey);
-            }
-            $this->services[$serviceKey] = $this->propagate($service);
+        if(is_object($service) &&  ($service instanceof ServiceLocatorAwareInterface)){
```
martijn-tao: There is no need to check if `$service` is an object. Just checking `$service instanceof ServiceLocatorAwareInterface` will give the desired result.

martijn-tao: Also to follow the PSR-2 standard the if keyword should have one space after it, and one space before the opening bracket. `if ($service instanceof ServiceLocatorAwareInterface) {`

common/oatbox/service/ServiceNotFoundException.php
```diff
@@ -21,12 +21,13 @@
 namespace oat\oatbox\service;
 
 use Zend\ServiceManager\Exception\ServiceNotFoundException as ZendException;
+use Psr\Container\NotFoundExceptionInterface;
 /**
```
martijn-tao: Following the PSR-2 guidelines there should be 1 newline after the `use` block

common/oatbox/service/ServiceNotFoundException.php
```diff
@@ -21,12 +21,13 @@
 namespace oat\oatbox\service;
 
 use Zend\ServiceManager\Exception\ServiceNotFoundException as ZendException;
+use Psr\Container\NotFoundExceptionInterface;
 /**
  * 
  * @author bout
  *
  */
```
martijn-tao: It might be a good idea to also update the docblock for this class

common/oatbox/service/ServiceManager.php
```diff
@@ -63,25 +68,69 @@ public function __construct($configService)
...
+
+    /**
+     * Try to initialize the class without parameters
+     *
+     * @param string $serviceId
+     * @param string $serviceKey
+     * @throws ServiceNotFoundException
+     * @return ConfigurableService
+     */
+    protected function tryAutowiring($serviceId, $serviceKey)
```
martijn-tao: Should we make this method private instead?

common/oatbox/service/ServiceManager.php
jbout: is_object check was added to ensure no  warnings are thrown here, and though it might not be required anymore I don't feel confident enough to remove it

PR #608, opened 2019-04-18
Fix/event manager and autowiring

siwane:
The goal of this PR is to be able to use autowired service as event listener. As discussed with Joel we are not implementing autowiring detection in service manager `has()` method but explicitly in eventmanager itself

### DIFF
--- a/test/unit/common/oatbox/event/EventManagerTest.php
+++ b/test/unit/common/oatbox/event/EventManagerTest.php
@@ -1,4 +1,22 @@
 <?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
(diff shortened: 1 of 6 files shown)


test/unit/common/oatbox/event/EventManagerTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public static $listened = false;
+
+    static public function listen()
+    {
+        self::$listened = true;
+    }
+}
+
+class FixtureEvent extends GenericEvent
```
Babacooll: Correct me if I'm wrong but we don't support PHP 5.6 anymore or do we ? Because if so, it would be cleaner to use anonymous classes here.

siwane: I unfortunately disagree with u, as our spielplatz is on 5.6 we have to be able to deploy to it.

jbout (approved):
Approve general approach

PR #616, opened 2019-05-15
Feature/mutex

hutnikau:


jbout:
merged but will require benchmarks before usage

PR #628, opened 2019-06-04
Fix/tao 7005/fix dbal sql mode

alexeyvy:
Make SQL_MODE for DBAL driver the same as PDO used to have


tikhanovichA:
can be reproduced only with mysql 5.7

PR #630, opened 2019-06-06
Use DBAL now expression

julien-sebire:
Requires https://github.com/oat-sa/lib-dbal-spanner/pull/2
Changing only the now expression to use dbal platform implementation.


julien-sebire:
Fixed in this commit: https://github.com/oat-sa/generis/commit/6521fb802ebcb022ef250e90dd427c087f1c2444#diff-ce3c63758a1a71ea1397fe4d4f1a4190

PR #640, opened 2019-06-26
Feature/nex 168/UUID primary keys

julien-sebire:
This trait allows generating a uuid value suitable for primary key for different entities, and is compatible with sharding feature on google spanner.


test/unit/helpers/UuidPrimaryKeyTraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace oat\generis\test\unit\helpers;
+
+use oat\generis\Helper\UuidPrimaryKeyTrait;
+use oat\generis\test\TestCase;
+
+class UuidPrimaryKeyTraitTest extends TestCase
+{
+    public function testgetUniquePrimaryKey()
```
emgolubev: I think there is a typo, should be `testGetUniquePrimaryKey`

helpers/UuidPrimaryKeyTrait.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
emgolubev: Just a question about all `PHP` files, should be there empty lines before and after license block?

test/unit/helpers/UuidPrimaryKeyTraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class UuidPrimaryKeyTraitUser
+{
+    use UuidPrimaryKeyTrait;
+}
+
+namespace oat\generis\Helper;
+
+function uniqid ($prefix = "", $more_entropy = false)
```
ekkinox: Instead of defining the namespaced function `uniqid()` yourself you could do that with php-mock-phpunit.

This has the advantage that you didn't create a side (potentially terrible) effect with your test, as the mock will be destroyed after a test.

see https://github.com/php-mock/php-mock#php-mock-api for example

test/unit/helpers/UuidPrimaryKeyTraitTest.php
julien-sebire: Correct. Fixed.

helpers/UuidPrimaryKeyTrait.php
julien-sebire: Not totally sure. In PSR-12, not yet approved, I can see this: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements
But I'm not sure it means we should have blank lines before license block. In fact, most of the files in find in tao code do not have a blank line before license block, so I'll remove it here. Thanks for seeing this, @emgolubev!

test/unit/helpers/UuidPrimaryKeyTraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public function testgetUniquePrimaryKey()
+    {
+        $subject = new UuidPrimaryKeyTraitUser();
+
+        $this->assertEquals('diuu desoppus a', $subject->getUniquePrimaryKey());
+    }
+}
+
+class UuidPrimaryKeyTraitUser
```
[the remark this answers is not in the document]
julien-sebire: Absolutely right. Don't know why I did it that way... :-) Fixed.

test/unit/helpers/UuidPrimaryKeyTraitTest.php
julien-sebire: Excellent! I was not aware of this one. Many thanks for this!

julien-sebire:
This uuid generator was moved to the persistence layer in which it should belong to.

ekkinox:
Imo, not all good. No problems to remove the trait if not needed but here you just merged an untested method to a core class...

PR #641, opened 2019-06-27
Feature/tao 8589/unique field validator

yuryfilippovich:
new ExtendedValidatorInterface for validators which need values another form fields to validate

### DIFF
--- a/common/oatbox/validator/ExtendedValidatorInterface.php
+++ b/common/oatbox/validator/ExtendedValidatorInterface.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


common/oatbox/validator/ExtendedValidatorInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+namespace oat\oatbox\validator;
+
+interface ExtendedValidatorInterface extends ValidatorInterface
+{
+    /**
+     * @param array $values
+     * @param string $currentPropertyUri
+     */
+    public function populateAdditionValues(array $values, $currentPropertyUri);
```
augustas: I would change the interface to
`public function populateWithFormValues(array $values, $currentPropertyUri);`
But in general I think the solution should be reviewed by an architect, to decide if this is the proper way to pass the modified resource id to a validator.

augustas (commented):
Looks okay, review by architect suggested

common/oatbox/validator/ExtendedValidatorInterface.php
zagovorichev: totally agree, we don't need this interface

zagovorichev (changes requested):

common/oatbox/validator/ExtendedValidatorInterface.php
yuryfilippovich: But if we need to validate some more complex form. For example with name1 and name2 and a least one field should be filled. In this case validation should be aware of other form values

PR #642, opened 2019-06-27
Feature/nex 168/UUID primary keys

julien-sebire:
This PR enables persistence layer to provide a uuid to be used as primary key.


julien-sebire:
Duplicates https://github.com/oat-sa/generis/pull/640.

PR #643, opened 2019-06-28
platform specific getNowExpression() is not usable.

julien-sebire:
Changed platform proxy to use platform specific date formatting function instead of getNowExpression.
We can't use $this->dbalPlatform->getNowExpression() because sqlite, used in the tests for instance, returns `datetime('now')` which can not be parsed as a regular date.
We instead generate a date with php and then format it with $this->dbalPlatform->getDateTimeTzFormatString(), to still have the correct format to be inserted in db depending on the platform.

### DIFF
--- a/common/persistence/sql/class.Platform.php
+++ b/common/persistence/sql/class.Platform.php
@@ -160,9 +160,7 @@ class common_persistence_sql_Platform {
         // correct format to be inserted in db.
 
         $datetime = new \DateTime('now', new \DateTimeZone('UTC'));
-        $date = $datetime->format($this->dbalPlatform->getDateTimeTzFormatString());
-
-        return $date;
+        return $datetime->format($this->dbalPlatform->getDateTimeTzFormatString());
     }
 
     /**
--- a/common/persistence/sql/class.Platform.php
+++ b/common/persistence/sql/class.Platform.php
@@ -160,7 +160,7 @@ class common_persistence_sql_Platform {
         // correct format to be inserted in db.
 
         $datetime = new \DateTime('now', new \DateTimeZone('UTC'));
-        return $datetime->format($this->dbalPlatform->getDateTimeFormatString());
+        return $datetime->format($this->getDateTimeFormatString());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


manifest.php
```diff
@@ -31,7 +31,7 @@
     'label' => 'Generis Core',
     'description' => 'Core extension, provide the low level framework and an API to manage ontologies',
     'license' => 'GPL-2.0',
-    'version' => '11.4.3',
+    'version' => '11.4.5',
```
peetya: Why not `11.4.4`?

common/persistence/sql/class.Platform.php
```diff
@@ -146,16 +146,41 @@ public function getMigrateSchemaSql($fromSchema,$toSchema){
...
-       return $date;
+        // We can't use $this->dbalPlatform->getNowExpression() because sqlite,
+        // at least used for the tests, returns `datetime('now')` which can
+        // not be parsed as a regular date.
+        // We instead generate a date with php and the format it with
+        // $this->dbalPlatform->getDateTimeTzFormatString(), to still have the
+        // correct format to be inserted in db.
+
+        $datetime = new \DateTime('now', new \DateTimeZone('UTC'));
+        $date = $datetime->format($this->dbalPlatform->getDateTimeTzFormatString());
```
peetya: How about `return $datetime->...;`?

julien-sebire: Fixed.

manifest.php
julien-sebire: Previously existing tag not merged...

common/persistence/sql/class.Platform.php
```diff
@@ -146,16 +146,39 @@ public function getMigrateSchemaSql($fromSchema,$toSchema){
...
+
+        $datetime = new \DateTime('now', new \DateTimeZone('UTC'));
+        return $datetime->format($this->dbalPlatform->getDateTimeTzFormatString());
+    }
+
+    /**
+     * Returns platform specific date formatting to store datetime field.
+     * @return string
+     */
+    public function getDateTimeFormatString()
```
bugalot: Is it in use? If not, let's not add additional code ;)

julien-sebire:
Added unit test for platform getNowExpression using platform specific datetime format.

PR #644, opened 2019-06-28
Feature/tao 8195/method not allowed param

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-8195

Optional param `$allowedMethods` added to `common_exception_MethodNotAllowed`.
It's needed to provide valid 405 response with `Allow` header later according to RFC (https://tools.ietf.org/html/rfc7231#section-6.5.5)

Needed for: https://github.com/oat-sa/tao-core/pull/2201

### DIFF
--- a/common/exception/class.MethodNotAllowed.php
+++ b/common/exception/class.MethodNotAllowed.php
@@ -25,6 +25,30 @@
  */
 class common_exception_MethodNotAllowed extends common_exception_BadRequest
 {
+    /**
+     * @var string[]|null
+     */
+    protected $allowedMethods;
+
+    /**
+     * @param string|null $message
+     * @param int $code
+     * @param string[]|null $allowedMethods
+     */
+    public function __construct($message = null, int $code = 0, $allowedMethods = null)
+    {
+        parent::__construct($message, $code);
+        $this->allowedMethods = $allowedMethods;
+    }
+
+    /**
+     * @return string[]|null
+     */
+    public function getAllowedMethods()
+    {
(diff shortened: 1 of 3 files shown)


common/exception/class.MethodNotAllowed.php
```diff
@@ -25,8 +25,32 @@
...
+     * @var string[]|null
+     */
+    protected $allowedMethods;
+
+    /**
+     * @param string|null $message
+     * @param int $code
+     * @param string[]|null $allowedMethods
+     */
+    public function __construct($message = null, int $code = 0, $allowedMethods = null)
```
hutnikau: Did we finally moved to php 7 (`int` type hint)?
What if restrict last parameter type to `array` and set default value `[]` ?

siarhei-baradzin: Thx, int type hint removed.

> What if restrict last parameter type to array and set default value [] ?

I have thought about it too, but decided to keep `string[]|null` to explicitly express that in case of `null` allowed methods are not specified by code which threw the exception in opposite to case when we have empty array which means empty list of allowed methods. Of course, the proper code shouldn't pass an empty array.

hutnikau: But why not restrict type to array? You still can use null default value, but when it passed, it should be an array:
```
public function __construct($message = null, int $code = 0, array $allowedMethods = null)
```

siarhei-baradzin: You are right, fixed

PR #646, opened 2019-07-03
Bumped minor version.

julien-sebire:
Bumped to a new minor version.


bugalot:
This happened because new methods were added in the last PR but only the bugfix version changed.

PR #652, opened 2019-07-16
Fix/annotations

jbout:


core/kernel/uri/MicrotimeRandUriProvider.php
```diff
@@ -43,10 +43,10 @@ class MicrotimeRandUriProvider extends ConfigurableService
     // --- OPERATIONS ---
     
     /**
-     * @return common_persistence_SqlPersistence
+     * @return \common_persistence_SqlPersistence
```
julien-sebire: Please use a `use` statement at the top of the file with an alias, e.g.:
```
use common_persistence_SqlPersistence as SqlPersistence;
...
@return SqlPersistence
```

core/kernel/uri/MicrotimeRandUriProvider.php
```diff
@@ -43,10 +43,10 @@ class MicrotimeRandUriProvider extends ConfigurableService
...
      */
     public function getPersistence() {
-        return \common_persistence_SqlPersistence::getPersistence($this->getOption(self::OPTION_PERSISTENCE));
+        return $this->getServiceLocator()->get(\common_persistence_Manager::SERVICE_ID)->getPersistenceById($this->getOption(self::OPTION_PERSISTENCE));
```
julien-sebire: Please use a `use` statement at the top of the file with an alias, e.g.:
```
use common_persistence_Manager as PersistenceManager;
...
@return PersistenceManager
```

core/kernel/uri/class.AdvKeyValueUriProvider.php
```diff
@@ -50,12 +51,10 @@ public function getPersistence() {
...
-     * @throws common_UriProviderException
+     * @throws UriProviderException
      */
     public function provide()
     {
-        $returnValue = (string) '';
-        
         $nextId = $this->getPersistence()->incr(self::PERSISTENCE_KEY);
         list($usec, $sec) = explode(" ", microtime());
         $uri = $this->getOption(self::OPTION_NAMESPACE) .'i'. (str_replace(".","",$sec."".$usec)) . $nextId;
```
julien-sebire: $uri is already a string so the casting on the return line is not necessary.

PR #654, opened 2019-07-22
Fix/human readable serialization

jbout:


common/class.Utils.php
```diff
@@ -181,6 +181,8 @@ public static function toHumanReadablePhpString($value, $indentation = 0)
                 }
             }
             $returnValue = "array(".substr($string, 0, -1).PHP_EOL.str_repeat('    ', $indentation).")";
+        } elseif (gettype($value) == "string") {
```
julien-sebire: It's better to use `is_string` function (and `is_array` above) since `gettype()` result strings are subject to change.

test/unit/UtilsTest.php
```diff
@@ -75,10 +75,14 @@ public function testSerialisation()
...
             'c' => array(
                 '1', '2', array(common_user_auth_Service::singleton())
-            )
+            ),
+            'd' => 'aaaaa'.PHP_EOL.'bbbbb'.PHP_EOL.'ccccc'
         );
         $value = eval("return " . common_Utils::toPHPVariableString($toSerialize) . ";");
         $this->assertEquals($toSerialize, $value);
+
+        $valueNice = eval("return " . common_Utils::toHumanReadablePhpString($toSerialize) . ";");
```
julien-sebire: Ouch, my eyes.... :-D
Why the `eval` here?
What about this ?
```
$valueNice = common_Utils::toHumanReadablePhpString($toSerialize);
```

jbout: We need to ensure that if the string is evaluated it will result in the same value as the original. How else to test this?

julien-sebire: You're right, here eval is used to ensure that the change doesn't affect the correctness of the modified code.

PR #657, opened 2019-07-29
Tao 8802/fix/adv key value persistence hset

hutnikau:
Current implementation of `common_persistence_InMemoryAdvKvDriver ::hset()` is wrong.
According to redis documentation: 
>1 if field is a new field in the hash and value was set.
>0 if field already exists in the hash and the value was updated.

### DIFF
--- a/common/persistence/class.InMemoryAdvKvDriver.php
+++ b/common/persistence/class.InMemoryAdvKvDriver.php
@@ -22,6 +22,14 @@ class common_persistence_InMemoryAdvKvDriver extends common_persistence_InMemory
 {
     const HPREFIX = 'hPrfx_';
 
+    /**
+     *
+     * @see common_persistence_Driver::connect()
+     */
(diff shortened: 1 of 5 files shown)


common/persistence/class.InMemoryAdvKvDriver.php
```diff
@@ -40,11 +48,9 @@ public function hExists($key, $field)
 
     public function hSet($key, $field, $value)
     {
-        if (! isset($this->persistence[$key])) {
-            return false;
-        }
+        $result = !isset($this->persistence[$key][self::HPREFIX . $field]);
```
zagovorichev: if you do not check that array key exists it could be a php warning with ba consequences

hutnikau: No, using `isset()` it's allowed to check existence of key in multidimensional array.

PR #676, opened 2019-09-30
Add type declaration for array parameter in methods of logger helper

Alroniks:
It fixes issues when somebody can pass a string instead of an array and it will break logger implementation that awaits array as a third parameter.

Changes done during work on https://oat-sa.atlassian.net/browse/TAO-9273 but not strongly required.


Alroniks:
These changes require a lot of attention as a logger used widely and it may bring unexpected behavior. So I am closing it for now as not relevant right now.

PR #678, opened 2019-10-09
TTD-18 Tests extends from `generis\test\TestCase`. All mock objects e…

Yury-Razhkou:
Related to: https://oat-sa.atlassian.net/browse/TTD-18
All tests extends `generis\test\TestCase` instead of `\PHPUnit_Framework_TestCase`
All Mock objects instances of `generis\test\MockObject` instead of `\PHPUnit_Framework_MockObject_MockObject`. Mocks was used only in phpdoc.
Possible way to test:
1) run unit tests.
2) update code.
3) run unit tests again. result should be the same as in 1.

### DIFF
--- a/test/TestCase.php
+++ b/test/TestCase.php
@@ -23,8 +23,10 @@ use Prophecy\Argument;
 use Zend\ServiceManager\ServiceLocatorInterface;
 use common_persistence_Manager;
 use common_persistence_sql_dbal_Driver;
+use PHPUnit\Framework\TestCase as UnitTestCase;
+use oat\generis\test\MockObject;
 
-abstract class TestCase extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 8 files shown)


test/TestCase.php
```diff
@@ -75,7 +77,7 @@ public function getSqlMock($key)
      * Returns a test double for the specified class.
      *
      * @param string $originalClassName
-     * @return \PHPUnit_Framework_MockObject_MockObject
+     * @return MockObject
      * @throws \PHPUnit_Framework_Exception
```
julien-sebire: Shall we do the same for \PHPUnit_Framework_Exception?

Yury-Razhkou: It is the only one occurrence of this class i found.

julien-sebire: Ok, so forget about it...

PR #679, opened 2019-10-10
Fix/system helper file upload limit

IgorSalevan:
Fixed calculation of upload file size limit. It's based on the composition of minimal value of 3 arguments. One of them is `php` setting `memory_limit` which could be equal to '-1'.
So we'll get an error on import operations.

### DIFF
--- a/helpers/SystemHelper.php
+++ b/helpers/SystemHelper.php
@@ -29,14 +29,17 @@ class SystemHelper
      */
     public static function getFileUploadLimit()
     {
+        $limits = [
+            self::toBytes(ini_get('upload_max_filesize')),
+            self::toBytes(ini_get('post_max_size'))
+        ];
 
-        $max_upload = self::toBytes(ini_get('upload_max_filesize'));
-        $max_post = self::toBytes(ini_get('post_max_size'));
-        $memory_limit = self::toBytes(ini_get('memory_limit'));
+        if (($memory_limit = self::toBytes(ini_get('memory_limit'))) !== -1) {
+            $limits[] = $memory_limit;
+        }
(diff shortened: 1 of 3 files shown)


helpers/SystemHelper.php
```diff
@@ -29,14 +29,17 @@ class SystemHelper
...
+            self::toBytes(ini_get('upload_max_filesize')),
+            self::toBytes(ini_get('post_max_size'))
+        ];
 
-        $max_upload = self::toBytes(ini_get('upload_max_filesize'));
-        $max_post = self::toBytes(ini_get('post_max_size'));
-        $memory_limit = self::toBytes(ini_get('memory_limit'));
-
-        $returnValue = min($max_upload, $max_post, $memory_limit);
+        if (($memory_limit = self::toBytes(ini_get('memory_limit'))) !== -1) {
```
chilly86: Please don't do value setting and value checking in the if!
Check value from the ini_get directly and in case it's not -1 just add the converted value to the limits

helpers/SystemHelper.php
```diff
@@ -29,14 +29,17 @@ class SystemHelper
...
+            $limits[] = $memory_limit;
+        }
+        $returnValue = min($limits);
 
-        return (int)$returnValue;
+        return $returnValue;
```
chilly86: you can return (int)min($limits) directly you don't need the variable

helpers/SystemHelper.php
IgorSalevan: Done.

helpers/SystemHelper.php
IgorSalevan: Done.

PR #683, opened 2019-10-15
Model id in namespace is not typed anymore.

julien-sebire:
Namespace can accept int or string as model id.

### DIFF
--- a/common/ext/class.Namespace.php
+++ b/common/ext/class.Namespace.php
@@ -40,7 +40,7 @@ class common_ext_Namespace
      * @access protected
      * @var int
      */
-    protected $modelId = 0;
+    protected $modelId;
 
     /**
      * the namespace URI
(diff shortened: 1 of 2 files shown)


common/ext/class.Namespace.php
```diff
@@ -40,130 +40,81 @@ class common_ext_Namespace
...
-        
-    	if($id > 0){
-    		$this->modelId = $id;
-    	}
-    	if(!empty($uri)){
-    		$this->uri = $uri;
-    	}
-    	
-        
+  		$this->modelId = $id;
```
mike-ionut-mihai-sandu-tao: should we check if `empty($id)` and use a uuid4 generator to have as default value ? 
also by removing the  check `if(!empty($uri)){` it can mean that the $uri can have the value `null`

common/ext/class.Namespace.php
```diff
@@ -40,130 +40,81 @@ class common_ext_Namespace
...
-        else{
-        	if (false === $db->exec("DELETE FROM models WHERE modelid = ?", array($this->getModelId()))){
-        		$returnValue = false;
-        	}
-        	else{
-        		$returnValue = true;
-        	}
+
+        // TODO refactor this to use triple store abstraction.
+        if (false === $db->exec("DELETE FROM models WHERE modelid = ?", array($this->getModelId()))){
```
mike-ionut-mihai-sandu-tao: the delete statement is usually dangerous if possible let's use something like a prepared statement

common/ext/class.Namespace.php
julien-sebire: It's only used for creating Namespace objects retrieved from db, so with either integer or strings as $id. It must not be called with null or no value...

jbout:
model table has been droped so this is no longer relevant

PR #686, opened 2019-10-15
User-modifiable model id is not hardcoded any more.

julien-sebire:
This PR makes the user-modifiable RDF model not hard-coded as "1" any more.


bugalot:
This cannot work without PR https://github.com/oat-sa/generis/pull/690/files (Abstract Model Factory)

jbout:
merged already

PR #694, opened 2019-10-24
Update UuidPrimaryKeyTrait.php

boajer:
UUID should be string instead of object to be persisted to storage


julien-sebire (approved):
Let's merge that.

PR #697, opened 2019-10-25
Support multiple log channels

peetya:
**Purpose**
The purpose of this modification is to support multiple logging channels in TAO in order to assign different handlers, processors and formatters to the different kind of logs. 

**Changes**
- Now it's possible to define multiple loggers in the configuration of `LoggerService`, using the `loggers` key, which expects an array of configurations (`logger` is still supported to ensure backward-compatibility)
- Channels can be defined with the `name` key for `TaoMonolog` logger. If no channel is defined, it will register itself for the default channel, which is `tao`
- Channel can be selected in the methods of `LoggerAwareTrait`. If it's not defined, it will send the logs to the default channel, which is `tao`.

**Example**
```
return new oat\oatbox\log\LoggerService([
    'loggers' => [
        [
            'class' => oat\oatbox\log\logger\TaoMonolog::class,
            'options' => [
                'name' => 'tao',
                'handlers' => [
                    [
                        'class' => Monolog\Handler\StreamHandler::class,
                        'options' => [
                            '/var/www/html/tao.log',
                        ],
                    ],
                ],
            ],
        ],
        [
            'class' => oat\oatbox\log\logger\TaoMonolog::class,
            'options' => [
                'name' => 'audit',
                'handlers' => [
                    [
                        'class' => Monolog\Handler\StreamHandler::class,
                        'options' => [
                            '/var/www/html/audit.log',
                        ],
                    ],
                ],
            ],
        ],
    ],
]);
```

**What to test**
- defining a single logger through `logger` key
- defining multiple loggers through `loggers` key
- defining multiple loggers for the same channel
- defining logger without channel
- logging to the default channel
- logging to a specific channel
- install TAO through a seed file, where single/multiple loggers defined

**TODO**
- [ ] Update documentation on TAO Hub


common/oatbox/log/LoggerService.php
```diff
@@ -29,98 +30,104 @@ class LoggerService extends ConfigurableService implements LoggerInterface
...
+            $this->registerLogger($logger, $channel);
+            return;
         }
 
-        if (!is_null($logger)) {
-            $this->logger = $logger;
-        } else {
-            $this->logger = new NullLogger();
+        if (!is_array($logger)) {
+            throw new \LogicException('Logger options must be an array');
```
boajer: nitpicking: shouldn't simplified NS be used here?

common/oatbox/log/LoggerService.php
```diff
@@ -29,98 +30,104 @@ class LoggerService extends ConfigurableService implements LoggerInterface
...
         }
 
-        return $this->logger;
+        if (!array_key_exists(self::OPTION_CLASS, $logger)) {
+            throw new \LogicException('No class defined for logger');
```
boajer: nitpicking: shouldn't simplified NS be used here?

common/oatbox/log/LoggerService.php
```diff
@@ -29,98 +30,104 @@ class LoggerService extends ConfigurableService implements LoggerInterface
...
+        }
+
+        if (!is_a($logger[self::OPTION_CLASS], LoggerInterface::class, true)) {
+            throw new \LogicException(sprintf('Logger class must implement %s', LoggerInterface::class));
```
boajer: nitpicking: shouldn't simplified NS be used here?

common/oatbox/log/logger/TaoMonolog.php
```diff
@@ -60,13 +62,23 @@ public function log($level, $message, array $context = array())
         $this->logger->log($level, $message, $context);
     }
 
+    public function getLogger(string $channel = null): LoggerInterface
```
boajer: Shouldn't class have consistent style -  some new methods are declared in PHP v7.0+ with scalar type declarations, etc, but older - no

common/oatbox/log/LoggerService.php
```diff
@@ -29,98 +30,104 @@ class LoggerService extends ConfigurableService implements LoggerInterface
...
+        if ($this->loggers === []) {
+            // To keep backward compatibility, "logger" key must be supported also
+            if ($this->hasOption(self::OPTION_LOGGER)) {
+                $this->loadLogger($this->getOption(self::OPTION_LOGGER));
+            }
+
+            $this->loadLoggers();
+        }
+
+        if (array_key_exists($channel, $this->loggers) && $this->loggers[$channel] instanceof LoggerInterface) {
```
siarhei-baradzin: you can remove multiple `$this->loggers[$channel]` access by key (3 times) by smth like:
```
return ($logger = $this->loggers[$channel] ?? null) && $logger instanceof LoggerInterface
    ? $logger
    : new NullLogger();
```

common/oatbox/log/LoggerService.php
```diff
@@ -29,98 +30,104 @@ class LoggerService extends ConfigurableService implements LoggerInterface
...
-     * Check if the logger is loaded from configuration
-     *
-     * @return bool
-     */
-    protected function isLoggerLoaded()
+    private function registerLogger(LoggerInterface $logger, string $channel): void
     {
-        return $this->logger instanceof LoggerInterface;
+        $this->loggers[$channel] = array_key_exists($channel, $this->loggers)
+            ? new LoggerAggregator([$logger, $this->loggers[$channel]])
```
siarhei-baradzin: it could cause nested `LoggerAggregator`s while it's possible to read list of loggers from existing `LoggerAggregator`, add new one to it and create new `LoggerAggregator` with flat array of loggers.

Keeping in mind that it's duplicated operation (see my commend above) consider to extract it to the separate method, for example located in `LoggerAggregator` itself:
```
/**
 * Creates new LoggerAggregator with all exising loggers and new $logger
 * @return LoggerAggregator
 */
public function withLogger(LoggerInterface $logger)
```

common/oatbox/log/logger/TaoMonolog.php
siarhei-baradzin: Agree, want to remind that PHPUnit 4.8 can't handle mocking classes with scalar type hinting

test/unit/oatbox/log/LoggerServiceTest.php
```diff
@@ -1,166 +1,168 @@
...
+{
+    /** @var LoggerService */
+    private $subject;
+
+    /** @var string */
+    private $logFilePath;
+
+    protected function setUp(): void
+    {
+        $this->logFilePath = tempnam(sys_get_temp_dir(), 'logtest');
```
siarhei-baradzin: This test looks more like an integration test that involves `SingleFileAppender` and writes data to a disk. As I understand, we should mock dependencies in unit tests, it will simplify it. `SingleFileAppender` should have own unit test.

boajer: Please consider https://github.com/oat-sa/tao-core/blob/master/test/TaoPhpUnitTestRunner.php#L141 as an abstraction for FS related operations for tests, if possible.

julien-sebire:
Out of date.

PR #700, opened 2019-11-08
Added check to ensure file is writeable before updating.

martijn-tao:
If a file can not be updated the call to `file_put_contents` causes a PHP warning. This warning is placed in the output buffer, and after this, there are several direct `header()` calls. This causes more warnings and causes the response to the front-end to be prepended with the PHP warnings (resulting in an unknown error).

In this PR a check is implemented to see if the file is indeed writable and if not a new exception is thrown.

### DIFF
--- a/common/oatbox/filesystem/File.php
+++ b/common/oatbox/filesystem/File.php
@@ -140,11 +140,15 @@ class File extends FileSystemHandler
     public function update($mixed, $mimeType = null)
     {
         if (!$this->exists()) {
-            throw new \FileNotFoundException('File "' . $this->getPrefix() . '" not found."');
+            throw new \FileNotFoundException(sprintf('File "%s" not found."', $this->getPrefix()));
+        }
+
+        if (!$this->isWriteable()) {
+            throw new FileNotWritableException(sprintf('Unable to write to file "%s"', $this->getPrefix()));
         }
 
         \common_Logger::i('Writting in ' . $this->getPrefix());
-        $config = (is_null($mimeType)) ? [] : ['ContentType' => $mimeType];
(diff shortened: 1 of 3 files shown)


common/oatbox/filesystem/utils/FileSystemWrapperTrait.php
```diff
@@ -243,6 +241,16 @@ public function addPlugin(PluginInterface $plugin) {
...
     }
 
+    /**
+     * Checks if a file exists and can be written to
+     *
+     * @param string $path
+     * @return bool
+     */
+    public function isWriteable($path) {
+        return is_writable($this->getFullPath($path));
```
siwane: Does it work with remote storage?

julien-sebire: Not existing in FlySystem... :(

julien-sebire (changes requested):
Hi Martĳn, do you think we can have some tests?

common/oatbox/filesystem/File.php
```diff
@@ -140,11 +140,15 @@ public function write($mixed, $mimeType = null)
     public function update($mixed, $mimeType = null)
     {
         if (!$this->exists()) {
-            throw new \FileNotFoundException('File "' . $this->getPrefix() . '" not found."');
+            throw new \FileNotFoundException(sprintf('File "%s" not found."', $this->getPrefix()));
+        }
+
+        if (!$this->isWriteable()) {
+            throw new FileNotWritableException(sprintf('Unable to write to file "%s"', $this->getPrefix()));
```
peetya: Missing test.

PR #703, opened 2019-12-04
Fix/tao 8584/unit test dependency propagation

jbout:


test/GenerisTestCase.php
```diff
@@ -59,18 +68,36 @@ protected function getOntologyMock()
...
         return $model;
     }
 
+    /**
+     * @return FileSystemService
+     */
+    protected function getFileSystemMock($dirs = [])
+    {
+        $adapterparam = [
+            'testfs' => class_exists('League\Flysystem\Memory\MemoryAdapter')
```
SashaTAO: Maybe it would be better to add `league/flysystem-memory` dependency in `require-dev` section in composer.json instead of checking if class exists?

PR #704, opened 2019-12-05
Fix/pass model rdfinsert

jbout:


ktsirkou:
Not needed anymore, will be on a future PR

PR #726, opened 2020-01-31
Feature/udi 21/added userservice interface

jbout:


common/oatbox/user/UserService.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param string $userId
+     * @return User
+     */
+    public function getUser($userId);
+
+    /**
+     * @param string[] $userIds
+     * @return User[]
+     */
+    public function getUsers($userIds);
```
SashaTAO: We should specify that parameter must be an array

PR #728, opened 2020-02-04
Add types to databse layer abstraction

siwane:
As preparation, for google spanner implementation, please find optional `$types` args added to persistence methods

### DIFF
--- a/common/persistence/class.SqlPersistence.php
+++ b/common/persistence/class.SqlPersistence.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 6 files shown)


common/persistence/sql/dbal/class.Driver.php
```diff
@@ -150,38 +155,40 @@ public function getSchemaManager()
     {
         return new common_persistence_sql_dbal_SchemaManager($this->connection->getSchemaManager());
     }
-    
-   
+
```
julien-sebire: Double blank line.

common/persistence/class.SqlPersistence.php
```diff
@@ -14,70 +15,79 @@
...
+
     /**
+     * Execute a statement.
+     *
+     * @param string $statement
+     * @param array $params
+     * @param array $types
+     * @return int number of updated rows
+     */
+    public function exec($statement, array $params = [], array $types = [])
```
mike-ionut-mihai-sandu-tao: do we need to alter and update any of the tests?

manifest.php
```diff
@@ -34,7 +34,7 @@
     'label' => 'Generis Core',
     'description' => 'Core extension, provide the low level framework and an API to manage ontologies',
     'license' => 'GPL-2.0',
-    'version' => '12.8.0',
+    'version' => '12.9.0',
```
julien-sebire: What about the version bump?

common/persistence/class.SqlPersistence.php
siwane: after checking nope, that may requires integration test

PR #731, opened 2020-02-05
Feature/nex 683/refactor modelid

siwane:


jbout (changes requested):
* Please do not rely on untestable deprecated core_kernel_api_ModelFactory
* Modelids can be reduced drasticly to 2 (read only and read/write model)
* We cannot change the default implementation until benchmark results are in

jbout:
No longer relevant as model table has been removed

PR #732, opened 2020-02-05
Feature/nex 683/clean code

siwane:
This PR has been created to extract code cleaning from spanner PR. Of course soen code has been touched and as it concerns persistence abstraction please find check list:

 - [x] test with mysql
- [x] test with postgres
- [x] test with serviceState on sqlkvdriver
- [x] test update with mysql
- [x] test update with psql

### DIFF
--- a/common/ext/class.Namespace.php
+++ b/common/ext/class.Namespace.php
@@ -108,10 +108,6 @@ class common_ext_Namespace
         }
 
         // TODO refactor this to use triple store abstraction.
-        if (false === $db->exec("DELETE FROM models WHERE modelid = ?", [$this->getModelId()])) {
-            return false;
-        }
-
-        return true;
+        return $db->exec("DELETE FROM models WHERE modelid = ?", [$this->getModelId()]);
     }
 }
--- a/common/ext/class.Namespace.php
+++ b/common/ext/class.Namespace.php
@@ -52,7 +52,7 @@ class common_ext_Namespace
      */
     public function __construct($id = '', $uri = '')
     {
-        $this->modelId = $id;
+        $this->modelId = (string) $id;
         $this->uri = $uri;
     }
 
--- a/common/ext/class.Namespace.php
+++ b/common/ext/class.Namespace.php
@@ -22,144 +22,96 @@
 /**
  * Short description of class common_ext_Namespace
  *
- * @access public
- * @author Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
+ * @access  public
+ * @author  Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
  * @package generis
-
  */
 class common_ext_Namespace
 {
-
-    // --- ASSOCIATIONS ---
-    // generateAssociationEnd :
-
-    // --- ATTRIBUTES ---
-
     /**
      * A unique identifier of the namespace
      *
-     * @access protected
-     * @var int
+     * @var string
      */
-    protected $modelId = 0;
+    protected $modelId;
 
     /**
      * the namespace URI
      *
-     * @access protected
      * @var string
      */
-    protected $uri = '';
-
-    // --- OPERATIONS ---
+    protected $uri;
 
     /**
-     * Create a namespace instance
+     * Namespace constructor.
      *
-     * @access public
      * @author Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
-     * @param  int id
+     *
+     * @param  string id
      * @param  string uri
-     * @return mixed
      */
-    public function __construct($id = 0, $uri = '')
+    public function __construct($id = '', $uri = '')
     {
-        
-        
-        if ($id > 0) {
-            $this->modelId = $id;
-        }
-        if (!empty($uri)) {
-            $this->uri = $uri;
-        }
+        $this->modelId = $id;
+        $this->uri = $uri;
     }
 
     /**
      * Get the identifier of the namespace instance
      *
-     * @access public
      * @author Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
-     * @return int
+     * @return string
      */
     public function getModelId()
     {
-        $returnValue = (int) 0;
-
-        
-        
-        $returnValue = $this->modelId;
-        
-        
-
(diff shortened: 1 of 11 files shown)


common/ext/class.Namespace.php
```diff
@@ -18,148 +18,100 @@
...
-            if (false === $db->exec("DELETE FROM models WHERE modelid = ?", [$this->getModelId()])) {
-                $returnValue = false;
-            } else {
-                $returnValue = true;
-            }
+            return false;
+        }
+
+        // TODO refactor this to use triple store abstraction.
+        if (false === $db->exec("DELETE FROM models WHERE modelid = ?", [$this->getModelId()])) {
```
mike-ionut-mihai-sandu-tao: I think here you can return directly the `$db->exec("DELETE FROM models WHERE modelid = ?", [$this->getModelId()])`
the issue i see is that from 2 times trying to delete the mode we are doing it now only once,
so please check this that we are not changing the scope of this code block

common/ext/class.Namespace.php
```diff
@@ -18,148 +18,100 @@
...
     }
-
+
     /**
      * Get the identifier of the namespace instance
      *
-     * @access public
      * @author Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
-     * @return int
+     * @return string
```
mike-ionut-mihai-sandu-tao: i don't think this will be true in case that the $this->modelId has received a int value

common/persistence/class.SqlKvDriver.php
```diff
@@ -76,30 +80,31 @@ public function set($id, $value, $ttl = null, $nx = false)
         try {
             $expire = is_null($ttl) ? 0 : time() + $ttl;
 
-            // we need int to have safe incr and decr methods
-            $encoded = is_int($value) ? $value : base64_encode($value);
-            $platformName = $this->sqlPeristence->getPlatForm()->getName();
+            $encoded = base64_encode($value);
```
boajer: We are loosing backward compatibility at that moment, right? Was that on purpose?

siwane: the purpose here is to unify value in KV, but agreed Im fixing backward compatibility

boajer: Unification is awesome, but that we need possibility to restore existing values properly, encoded in the old-fashion

common/persistence/class.SqlKvDriver.php
```diff
@@ -119,20 +125,19 @@ public function get($id)
...
-            $sessionValue = $this->sqlPeristence->query($statement, [$id]);
+            $statement = $this->sqlPersistence->getPlatForm()->limitStatement($statement, 1);
+            $sessionValue = $this->sqlPersistence->query($statement, [$id]);
             while ($row = $sessionValue->fetch()) {
-                if ($row['kv_time'] == 0 || $row['kv_time'] >= time()) {
-                    return filter_var($row['kv_value'], FILTER_VALIDATE_INT) !== false
-                        ? (int)$row['kv_value']
-                        : base64_decode($row['kv_value']);
+                if ($row["kv_time"] == 0 || $row["kv_time"] >= time()) {
+                    return base64_decode($row["kv_value"]);
```
boajer: please see https://github.com/oat-sa/generis/pull/732/files#r375290983

common/ext/class.Namespace.php
siwane: the behaviour is the same as before. I have factorize the code to return directly exec ouput

common/persistence/class.SqlKvDriver.php
siwane: I'm keeping this part to be able to restore old int value, new ones will be base64encoded

common/persistence/class.SqlKvDriver.php
siwane: I'm keeping this behaviour but at decoding, Im taking care of old fashion way

common/ext/class.Namespace.php
julien-sebire: Maybe we should cast to string in the constructor.

common/ext/class.Namespace.php
julien-sebire: No, this first removes every **statement** linked to a model id, then the **model** itself.

common/persistence/class.SqlKvDriver.php
```diff
@@ -76,30 +80,31 @@ public function set($id, $value, $ttl = null, $nx = false)
...
                 $statement = "MERGE INTO kv_store USING DUAL ON(kv_id = :id) 
                     WHEN NOT MATCHED THEN INSERT (kv_id, kv_value, kv_time) VALUES (:id, :data, sysdate) 
                     WHEN MATHED THEN UPDATE SET kv_value = :data WHERE kv_id = :id";
             } else {
                 $statement = 'UPDATE kv_store SET kv_value = :data , kv_time = :time WHERE kv_id = :id';
-                $returnValue = $this->sqlPeristence->exec($statement, $params);
-                if ($returnValue == 0) {
-                    $returnValue = $this->sqlPeristence->insert('kv_store', ['kv_id' => $id, 'kv_time' => $expire, 'kv_value' => $encoded]);
+                $returnValue = $this->sqlPersistence->exec($statement, $params, ['data' => ParameterType::STRING, 'time' => ParameterType::INTEGER, 'id' => ParameterType::STRING]);
+                if (0 === $returnValue) {
```
julien-sebire: No yoda condition please. ;-)

common/ext/class.Namespace.php
```diff
@@ -18,148 +18,100 @@
...
-     * @access public
      * @author Jerome Bogaerts, <jerome.bogaerts@tudor.lu>
-     * @param  int id
+     *
+     * @param  string id
      * @param  string uri
-     * @return mixed
      */
-    public function __construct($id = 0, $uri = '')
+    public function __construct($id = '', $uri = '')
```
julien-sebire: Can be changed to
`public function __construct($id = '', $uri = '')`
with no problem.

mike-ionut-mihai-sandu-tao:

PR #734, opened 2020-02-17
Bump version

siwane:
The purpose of this PR is to restore the done by https://github.com/oat-sa/generis/pull/714.
This is a fix for the bug described in the PR comment

### DIFF
--- a/common/persistence/class.SqlKvDriver.php
+++ b/common/persistence/class.SqlKvDriver.php
@@ -24,6 +24,7 @@ use Doctrine\DBAL\ParameterType;
 /**
  * A key value driver based upon an existing sql persistence
  *
+ * @todo : Refactor driver specific stuff to dedicated implementation
  */
 class common_persistence_SqlKvDriver implements common_persistence_KvDriver
 {
@@ -80,7 +81,8 @@ class common_persistence_SqlKvDriver implements common_persistence_KvDriver
         try {
             $expire = is_null($ttl) ? 0 : time() + $ttl;
 
-            $encoded = base64_encode($value);
+            // we need int to have safe incr and decr methods
+            $encoded = is_int($value) ? $value : base64_encode($value);
 
             $platformName = $this->sqlPersistence->getPlatForm()->getName();
             $params = [':data' => $encoded, ':time' => $expire, ':id' => $id];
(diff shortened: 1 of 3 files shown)


siwane:
As discussed, Incr/decr for sqlkv will be done at php level to be compliant with postgres and spanner

common/persistence/class.SqlKvDriver.php
```diff
@@ -181,25 +182,33 @@ public function del($id)
...
+     * @return boolean
+     *
+     * @throws common_Exception
      */
     public function decr($id)
     {
-        $params = [':id' => $id];
-        $statement = 'UPDATE kv_store SET kv_value = kv_value - 1 WHERE kv_id = :id';
-        return $this->sqlPersistence->exec($statement, $params);
+        $value = filter_var($this->get($id), FILTER_VALIDATE_INT);
```
zagovorichev: Example:
try to run 1000 test takers at the same time I'm pretty sure that it will lead us to the situation when one updater will overwrite another value, I mean that when 1 php process gets value - another parallel process gets the same value, so on update these processes will write the same incorrect value.
1.0 get 10 delivery executions
1.1 get 10 de
2.0 10-1 write 9 to db
2.1 write 9 to db as well, but there should be 8 already

zagovorichev (changes requested):

common/persistence/class.SqlKvDriver.php
```diff
@@ -181,25 +182,33 @@ public function del($id)
...
+     * @return boolean
+     *
+     * @throws common_Exception
      */
     public function incr($id)
     {
-        $params = [':id' => $id];
-        $statement = 'UPDATE kv_store SET kv_value = kv_value + 1 WHERE kv_id = :id';
-        return $this->sqlPersistence->exec($statement, $params);
+        $value = filter_var($this->get($id), FILTER_VALIDATE_INT);
```
boajer: Isn't that risky without transactions?

common/persistence/class.SqlKvDriver.php
siwane: agreed, will revert to previous implementation and add spanner flavor

common/persistence/class.SqlKvDriver.php
siwane: as discussed with @zagovorichev , Im going to use update query instead of get/set

common/persistence/class.SqlKvDriver.php
```diff
@@ -181,24 +184,42 @@ public function del($id)
     /**
      * Increment existing value
      * @param string $id
-     * @return mixed
+     * @return boolean
```
boajer: _common_persistence_SqlPersistence::exec_ returns  _int_  but not boolean

common/persistence/class.SqlKvDriver.php
```diff
@@ -181,24 +184,42 @@ public function del($id)
...
         $params = [':id' => $id];
-        $statement = 'UPDATE kv_store SET kv_value = kv_value + 1 WHERE kv_id = :id';
         return $this->sqlPersistence->exec($statement, $params);
     }
 
     /**
      * Decrement existing value
      * @param $id
-     * @return mixed
+     * @return boolean
```
boajer: common_persistence_SqlPersistence::exec returns int but not boolean

boajer (changes requested):
Minor chances needed in PHPDoc

common/persistence/class.SqlKvDriver.php
```diff
@@ -181,24 +184,42 @@ public function del($id)
...
      */
     public function decr($id)
     {
+        switch ($this->sqlPersistence->getPlatForm()->getName()) {
+            case 'psql':
```
zagovorichev: I've gotten `'postgresql'` platform name
(and it was:         `$intVal = $platformName == 'postgresql' ? 'kv_value::integer' : 'kv_value';` in previous implementation)

zagovorichev (changes requested):

common/persistence/class.SqlKvDriver.php
siwane: 😭 blame on me

zagovorichev (approved):

boajer (approved):
LGFM

mike-ionut-mihai-sandu-tao (approved):

PR #737, opened 2020-02-18
Feature/nex 683/rmove model table

siwane:
- `core_kernel_api_ModelFactory` has been refactored to use default SmootModel value instead of model table
- It works at update, but after fresh install only 2 models readable exists

### DIFF
--- a/core/kernel/api/class.ModelFactory.php
+++ b/core/kernel/api/class.ModelFactory.php
@@ -28,7 +28,6 @@ use core_kernel_persistence_smoothsql_SmoothModel as SmoothModel;
 class core_kernel_api_ModelFactory
 {
     /**
-     * @author "Lionel Lecaque, <lionel@taotesting.com>"
      * @param string $namespace
      * @param string $data xml content
      */
--- a/core/kernel/api/class.ModelFactory.php
+++ b/core/kernel/api/class.ModelFactory.php
@@ -34,7 +34,7 @@ class core_kernel_api_ModelFactory
      */
     public function createModel($namespace, $data)
     {
-        $modelId = SmoothModel::DEFAULT_READABLE_MODEL;
+        $modelId = SmoothModel::DEFAULT_READ_ONLY_MODEL;
 
         $modelDefinition = new EasyRdf_Graph($namespace);
         if (is_file($data)) {
--- a/core/kernel/api/class.ModelFactory.php
+++ b/core/kernel/api/class.ModelFactory.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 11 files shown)


core/kernel/api/class.ModelFactory.php
```diff
@@ -14,58 +15,27 @@
...
-     * @param string $namespace
-     */
-    private function addNewModel($namespace)
-    {
-        $dbWrapper = core_kernel_classes_DbWrapper::singleton();
-        $results = $dbWrapper->insert('models', ['modeluri' => $namespace]);
-    }
-    
     /**
      * @author "Lionel Lecaque, <lionel@taotesting.com>"
```
julien-sebire: Unnecessary @author annotation.

core/kernel/persistence/smoothsql/install/SmoothRdsModel.php
```diff
@@ -36,11 +36,11 @@ class SmoothRdsModel
...
-        $table = $schema->createTable("models");
-        $table->addColumn('modelid', "integer", ["notnull" => true,"autoincrement" => true]);
-        $table->addColumn('modeluri', "string", ["length" => 255,"default" => null]);
-        $table->addOption('engine', 'MyISAM');
-        $table->setPrimaryKey(['modelid']);
+//        $table = $schema->createTable("models");
+//        $table->addColumn('modelid', "integer", ["notnull" => true,"autoincrement" => true]);
+//        $table->addColumn('modeluri', "string", ["length" => 255,"default" => null]);
+//        $table->addOption('engine', 'MyISAM');
+//        $table->setPrimaryKey(['modelid']);
```
julien-sebire: Please remove the commented out code.

siwane:
@julien this is still under construction as discussed with joel, thanks for comments. Have u seen something strange statements table around model id field?

core/kernel/persistence/smoothsql/install/SmoothRdsModel.php
siwane: yes, will do, it is for highlight the removal

julien:
@siwane no sorry, I don't know what you're talking about

core/kernel/persistence/smoothsql/install/SmoothRdsModel.php
jbout: if the table is no longer created, should we not drop it during updates as well?

siwane:
The pr is about deleting model table, because as you may remember, it contains autoincrement. As discussed with joel, the table is not needed anymore. But it can implies some side effect, have u found some?

core/kernel/persistence/smoothsql/class.SmoothRdf.php
```diff
@@ -65,7 +66,7 @@ public function get($subject, $predicate)
     public function add(\core_kernel_classes_Triple $triple)
     {
         if (!in_array($triple->modelid, $this->model->getReadableModels())) {
-            $this->model->addReadableModel($triple->modelid);
+            $triple->modelid = SmoothModel::DEFAULT_READ_ONLY_MODEL;
```
jbout: Why would this be required?

core/kernel/persistence/smoothsql/install/SmoothRdsModel.php
siwane: I have some doubt about deleting table during update, do u think we can remove it?

PR #744, opened 2020-03-05
Fix/tdr 1/testcaase must be compatible

bartlomiejmarszal:
In order to work with PHPUnit `7.0.3` I made `TestCase` class that is extending `PHPUnit\Framework\TestCase` compatible ith new PHPunit version.


test/TestCase.php
```diff
@@ -26,18 +26,10 @@
 use common_persistence_Manager;
 use common_persistence_sql_dbal_Driver;
 use PHPUnit\Framework\TestCase as UnitTestCase;
-use oat\generis\test\MockObject;
+use PHPUnit\Framework\MockObject\MockObject;
```
julien-sebire: If you make this change, I think you can then remove the `oat\generis\test\PHPUnit_Framework_MockObject_MockObject` class.

PR #754, opened 2020-03-19
Added a method to test the inclusion of an absolute path in another.

julien-sebire:
This PR adds a method to check if an absolute path is contained within another absolute given path.

Required by https://github.com/oat-sa/extension-tao-itemqti/pull/1411

### DIFF
--- a/helpers/class.File.php
+++ b/helpers/class.File.php
@@ -104,16 +104,7 @@ class helpers_File
      */
     public static function isFileInsideDirectory($filename, $directory)
     {
-        $canonicalDirectory = realpath($directory);
-        if (false === $canonicalDirectory) {
-            return false;
-        }
-        $canonicalFilename = realpath($canonicalDirectory . DIRECTORY_SEPARATOR . $filename);
-        if (false === $canonicalFilename) {
-            return false;
-        }
-
-        return 0 === strpos($canonicalFilename, $canonicalDirectory);
+        return self::isAbsoluteFileInsideDirectory($directory . DIRECTORY_SEPARATOR . $filename, $directory);
     }
 
     /**
--- a/helpers/class.File.php
+++ b/helpers/class.File.php
@@ -98,8 +98,8 @@ class helpers_File
     /**
      * Helps prevent 'path traversal' attacks
      *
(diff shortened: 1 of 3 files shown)


helpers/class.File.php
```diff
@@ -112,6 +112,28 @@ public static function isFileInsideDirectory($filename, $directory)
...
+    }
+
+    /**
+     * Helps prevent 'path traversal' attacks
+     *
+     * @param string $filename  absolute path to file
+     * @param string $directory absolute path to directory
+     * @return bool
+     */
+    public static function isAbsoluteFileInsideDirectory($filename, $directory)
```
wazelin: I'd ask to add some test cases for the method, but I realize php-vfs won't be able to help you with `realpath`, and the helper is in the global namespace, so you won't even be able to mock the entire function :/

ktsirkou: What if we use the previous method `isFileInsideDirectory` to validate  file despite if is absolute path or not?
I think a check like 
```
canonicalFilename = realpath( $filename );

if ( canonicalFilename !== $filename ) {
         $canonicalFilename = realpath($canonicalDirectory . DIRECTORY_SEPARATOR . $filename);
    }
```
What do you think?

julien-sebire: The reason why I added this is because originally, the tested file structure was this one:
```
path
`-- to
    `-- uncompressed package.zip
        |-- item1
        |   `-- item1.xml
        `-- shared
            `-- shared stimulus.xml
```
with the code in `item1.xml` was including `../shared/shared-stimulus.xml`, and the test was actually checking if `/path/to/uncompressed package.zip/item1/../shared/shared-stimulus.xml` was actually inside '/path/to/uncompressed package.zip/item1/', which is obviously not the case.
The idea was to prevent exposing the whole filesystem to the user.

What I did was to check if the absolute `/path/to/uncompressed package.zip/shared/shared stimulus.xml` is actually inside `/path/to/uncompressed package.zip/`.

julien-sebire: I totally agree. I'll figure out what to do. Probably integration test then.

PR #757, opened 2020-03-23
Merge/nex 778/fix lib generis search

bartlomiejmarszal:
After we update lib-generis-search we did not update this dependency in generis

https://oat-sa.atlassian.net/browse/NEX-778


julien-sebire (changes requested):
Why would you request a fix version which is already covered by the `^2.1.0` constraint?
`^2.1.0` means `>=2.1.0 && < 2.2`

bartlomiejmarszal:
You are right. I'm closing this PR due to:
https://github.com/oat-sa/generis/pull/758

PR #762, opened 2020-04-02
Cleanup/manifest

jbout:


config/default/LockService.conf.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+/**
```
boajer, line 3: should we have headers for the default configs stored separately?

config/default/LockService.conf.php
```diff
@@ -0,0 +1,9 @@
...
+ * Default config header created during install
+ */
+
+return new oat\oatbox\mutex\LockService(array(
```
boajer, line 7: what about shor array notation?

config/default/LockService.conf.php
```diff
@@ -0,0 +1,9 @@
...
+    'persistence_class' => 'oat\\oatbox\\mutex\\NoLockStorage'
```
boajer, line 8: Any reason of usage literals but not constants?

manifest.php
```diff
@@ -23,6 +23,7 @@
  */
 
 use oat\generis\scripts\install\SetupDefaultKvPersistence;
+use oat\generis\scripts\install\TaskQueue;
```
boajer, line 26: Should we change copyright year at the very top of this file to 2020?

gabrielfs7: off: How you generally keep the copyright of the files updated? Only when there is a file modification? If we do not have an automatic way to do it, maybe it should be included in the Michael's checklist.

PR #763, opened 2020-04-03
Feature/replicate item tree structure into media manager

julien-sebire:
This PR adds methods to Rdf Class resource to:
- retrieve or create a class by label in the direct subclasses of a class,
- create several levels of tree structure within one call.

Required by https://github.com/oat-sa/extension-tao-mediamanager/pull/151


core/kernel/classes/class.Class.php
```diff
@@ -228,6 +228,52 @@ public function createSubClass($label = '', $comment = '', $uri = "")
         return $returnValue;
     }
 
+    /**
+     * Retrieves a direct subclass by label.
+     *
+     * @param  string label
+     * @return core_kernel_classes_Class|null
+     */
+    public function retrieveSubClassByLabel($label)
```
SergiiTao, line 237: can we set type hints and a return type here and below?

core/kernel/classes/class.Class.php
```diff
@@ -228,6 +228,52 @@ public function createSubClass($label = '', $comment = '', $uri = "")
...
+     * @param  array $labels indexed array of labels ordered from root to leaf class
+     * @return core_kernel_classes_Class The last class created
+     */
+    public function createSubClassPathByLabel(array $labels)
+    {
+        $currentClass = $this;
+
+        foreach($labels as $label) {
+            $currentClass = $currentClass->retrieveOrCreateSubClassByLabel($label);
+        }
```
SergiiTao, line 272: I don't think we need to go through all array `$labels`
seems it is enough to `retrieveOrCreateSubClassByLabel` only the last item in the array

wazelin: Not really, `createClass`, if it occurs along the path, triggers `ResourceCreated`. So all sorts of side effects can happen in its handlers.

julien-sebire: We have to do it for each level of the tree, since some (at the beginning) can already exist. We then retrieve the existing ones and create the non-existing ones.
`retrieveOrCreateSubClassByLabel` only creates one subclass.

core/kernel/classes/class.Class.php
julien-sebire: Sure. I don't know the current Tao policy regarding this, and I kept the style of the rest of the file... You tell me. ;-)

julien-sebire:
> do we have a ticket for that?

I'd suggest BOSA-5 counterpart in TAO? @KRAMAOAT  or @ktsirkou can you think about that?

julien-sebire:
> Can you exaplain your approach in test related to creating a mock model there?

The idea is to make a unit test, without requiring the configuration from an installed instance of Tao. That would make it an integration test, and we don't run nor maintain integration tests in Tao...

core/kernel/classes/class.Class.php
```diff
@@ -228,6 +228,52 @@ public function createSubClass($label = '', $comment = '', $uri = "")
...
+    /**
+     * Retrieves a direct subclass by label or creates it if not existent.
+     *
+     * @param  string label
+     * @return core_kernel_classes_Class
+     */
+    public function retrieveOrCreateSubClassByLabel($label)
+    {
+        return $this->retrieveSubClassByLabel($label) ?: $this->createSubClass($label);
+    }
```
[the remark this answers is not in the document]
julien-sebire: I think that `retrieveOrCreate` is more self-explanatory than `get` in here... Opinion?

core/kernel/classes/class.Class.php
```diff
@@ -228,6 +228,52 @@ public function createSubClass($label = '', $comment = '', $uri = "")
...
+
+    /**
+     * Retrieves a direct subclass by label or creates it if not existent.
+     *
+     * @param  string label
+     * @return core_kernel_classes_Class
+     */
+    public function retrieveOrCreateSubClassByLabel($label)
+    {
+        return $this->retrieveSubClassByLabel($label) ?: $this->createSubClass($label);
```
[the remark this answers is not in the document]
julien-sebire: No, I'm creating a subclass if it doesn't exist. I read it this way:
If I can retrieve a subclass with the given label, I return it, else I create a subclass with the given label and return it.

test/unit/core/kernel/classes/ClassTest.php
```diff
@@ -0,0 +1,211 @@
...
+
+        $newLabel = 'a brand new label';
+
+        $this->assertEquals(
+            $this->createClassWithLabel($newLabel),
+            $this->subject->createSubClassPathByLabel([$label, $newLabel])
+        );
+    }
+
+    private function createModel(array $subClassLabels, $triggeredEvents = 0)
```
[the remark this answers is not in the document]
julien-sebire: Yes, I've seen this done in different places like https://github.com/oat-sa/tao-core/blob/master/test/unit/model/taskQueue/Worker/AbstractWorkerTest.php#L139 for example. Is there a standardized way to do it?

core/kernel/classes/class.Class.php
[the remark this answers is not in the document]
wazelin: Agree, either that or `fetch`.
`get` is misleading as it's mostly represents an accessor method.

julien-sebire:
> do we have a ticket for that?

It seems we don't need a ticket on this since it's product-related...

SergiiTao:
> > do we have a ticket for that?
> 
> It seems we don't need a ticket on this since it's product-related...

we need to log our hours and define a task what we are working on

ktsirkou:
> > do we have a ticket for that?
> 
> I'd suggest BOSA-5 counterpart in TAO? @KRAMAOAT or @ktsirkou can you think about that?

BOSA-5 is an epic that is only related to the migration preparation as I see. 
Creating a ticket is no that problematic 
let me create it

ktsirkou:
> > > do we have a ticket for that?
> > 
> > 
> > It seems we don't need a ticket on this since it's product-related...
> 
> we need to log our hours and define a task what we are working on

here is the ticket https://oat-sa.atlassian.net/browse/BSA-100

PR #765, opened 2020-04-09
Added basic session context support

jbout:
Currently in order to add context information to a session we have to extend the session itself (see taoLti). This feature would allow us to add different kind of contexts to the basic session

### DIFF
--- a/common/oatbox/session/SessionContext.php
+++ b/common/oatbox/session/SessionContext.php
@@ -24,7 +24,7 @@ declare(strict_types=1);
 namespace oat\oatbox\session;
 
 /**
- * Represents an additional context that can be attached to a sesson
+ * Represents an additional context that can be attached to a session
  */
 interface SessionContext
 {
--- a/common/oatbox/session/SessionContext.php
+++ b/common/oatbox/session/SessionContext.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+declare(strict_types=1);
+
+namespace oat\oatbox\session;
+
+/**
+ * Represents an additional context that can be attached to a sesson
+ */
+interface SessionContext
+{
+}
--- a/common/session/class.AnonymousSession.php
+++ b/common/session/class.AnonymousSession.php
@@ -33,26 +33,12 @@ use oat\oatbox\user\UserLanguageServiceInterface;
  * @package generis
 
  */
-class common_session_AnonymousSession implements common_session_StatelessSession, ServiceLocatorAwareInterface
+class common_session_AnonymousSession extends common_session_BasicSession implements common_session_StatelessSession
 {
-    use ServiceLocatorAwareTrait;
-
-    /**
-     * (non-PHPdoc)
-     * @see common_session_Session::getUser()
-     */
-    public function getUser()
-    {
-        return new AnonymousUser();
-    }
     
-    /**
-     * (non-PHPdoc)
-     * @see common_session_Session::getUserUri()
-     */
-    public function getUserUri()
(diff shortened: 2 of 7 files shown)


common/oatbox/session/SessionContext.php
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+declare(strict_types=1);
+
+namespace oat\oatbox\session;
+
+/**
+ * Represents an additional context that can be attached to a sesson
```
bartlomiejmarszal, line 27:
```suggestion
 * Represents an additional context that can be attached to a session
```

common/session/class.BasicSession.php
```diff
@@ -165,4 +168,20 @@ public function setServiceLocator(ServiceLocatorInterface $serviceLocator)
...
         return $this->setOriginalServiceLocator($serviceLocator);
     }
+
+    public function withContext(SessionContext $context): self
+    {
+        $this->contexts[] = $context;
+        return $this;
+    }
+
+    public function getContexts($class = null): array
```
bartlomiejmarszal, line 178: The method does not have any tests

gabrielfs7, line 178:
```suggestion
    public function getContexts(string $class = null): array
```

common/session/class.BasicSession.php
```diff
@@ -165,4 +168,20 @@ public function setServiceLocator(ServiceLocatorInterface $serviceLocator)
...
+    {
+        $contexts = $this->contexts;
+        if ($class != null) {
+            $contexts = array_filter($contexts, function($element) use ($class) {return $element instanceof $class;});
+        }
+        return $contexts;
```
gabrielfs7, line 184: I would use **early return** as it is recommended to avoid usage of `else` or unnecessary pre-assignments (https://williamdurand.fr/2013/06/03/object-calisthenics/#2-dont-use-the-else-keyword).

 ```suggestion
        if ($class === null) {
            return $this->contexts;
        }
        
        return array_filter(
            $this->contexts, 
            function($element) use ($class) {
                return $element instanceof $class;
            }
        );
```
You can avoid creating another variable unnecessarily. I also think we should indent anonymous functions for better readability.

Comment (jbout) [reply common/session/class.BasicSession.php:184]: While in general the Object Calisthenics are a good read, I strongly disagree with the early return. While early returns are acceptable as part of guards, they should be avoided as they are easy to miss and make the flow of a function harder to understand.

The article in question was about avoiding **ELSE** statements and even suggested to introduce a variable in order to make your return statement parametrizable, as I did.

File: common/session/class.BasicSession.php (hunk 2 above)
Comment (jbout) [reply common/session/class.BasicSession.php]: fixed

File: common/oatbox/session/SessionContext.php (hunk 1 above)
Comment (jbout) [reply common/oatbox/session/SessionContext.php]: fixed

File: common/session/class.BasicSession.php (hunk 2 above)
Comment (jbout) [reply common/session/class.BasicSession.php]: added simple test

File: common/session/class.AnonymousSession.php [hunk 4]
```diff
@@ -120,4 +120,10 @@ public function refresh()
     {
         // nothing to do here
     }
+
+    public function getContexts(string $class = null): array
+    {
+        return [];
```
boajer, line 126: Why should we drop contexts support for anonymous?

jbout: good idea, added support

test/unit/common/session/BasicSessionTest.php
```diff
@@ -0,0 +1,46 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ *
```
boajer, line 19: Looks like more empty lines that should be in a header

jbout: fixed

common/session/class.BasicSession.php
gabrielfs7: I am recommending object-calisthenics cause they are well know and respected recommendations. I agree with early return, since it will quickly avoid me to continue reading the code when I reached the point I was looking for. It saves time and it is more fluid. 

I would at least consider indenting the anonymous function to make code easier to be read by other colleagues.

common/oatbox/session/SessionContext.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\oatbox\session;
+
+/**
+ * Represents an additional context that can be attached to a session
+ */
+interface SessionContext
```
gabrielfs7, line 29: Shouldn't this interface have some behavior defined? 

As a context it should at least have some contextData (_getContextData(): array_ and _addContextData(string $id, $data): void_) otherwise this interface will force user to break `open/closed` principle as the client of this context will have to be aware of implementation instead or relying in the interface.

PR #766, opened 2020-04-09
Feature/tbd 18/di improvement

SergiiTao:
explanation here oat-sa/tao-core#2480

### DIFF
--- a/common/oatbox/service/ServiceFactoryInterface.php
+++ b/common/oatbox/service/ServiceFactoryInterface.php
@@ -25,7 +25,7 @@ namespace oat\oatbox\service;
 
 use Zend\ServiceManager\ServiceLocatorInterface;
 
-interface ServiceFactory
+interface ServiceFactoryInterface
 {
     public function __invoke(ServiceLocatorInterface $serviceLocator);
 }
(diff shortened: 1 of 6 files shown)


common/oatbox/service/ServiceManager.php
```diff
@@ -108,6 +108,11 @@ private function load($serviceId, $serviceKey)
         if ($service === false) {
             $service = $this->tryAutowiring($serviceId, $serviceKey);
         }
+
+        if ($service instanceof ServiceFactoryInterface) {
+            $service = $service($this);
```
wazelin, line 113: Aha, so you register a factory, but when you call the service locator with its ID, you're gonna receive an actual service, made by that factory. Tricky, and may potentially be very misleading.

common/oatbox/service/ServiceFactoryInterface.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
```
wazelin, line 3: Declare statements should be placed below the file-level docbloc, BTW.

https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements

PR #771, opened 2020-04-16
Update composer.json

boajer:


Alroniks:
I am closing it to reduce a mess as support of composer 2.0 is successfully implemented already.

PR #777, opened 2020-04-29
Remove __serialize method in basic Container class

Alroniks:
This method became magic in PHP 7.4 and current implementation raises fatal issues.

If you are running tao on 7.4 version, attempt to launch delivery will fail due to such error message
```
Warning: session_start(): Failed to decode session object. Session has been destroyed in /var/www/html/tao/models/classes/mvc/Bootstrap.php on line 292
```
I tried only LTI launch, but I guess it will fail on regular too.

### DIFF
--- a/core/kernel/classes/class.Container.php
+++ b/core/kernel/classes/class.Container.php
@@ -30,17 +30,4 @@
  */
 class core_kernel_classes_Container extends common_Object
 {
-
-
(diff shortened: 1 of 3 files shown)


Alroniks:
Closing it because it was fixed here https://github.com/oat-sa/generis/pull/799

PR #785, opened 2020-05-06
Post update hook in updater scripts

hutnikau:
PR implement post update hooks for all the extensions. 
It may be used for warming up the cache, synchronizing data, etc.

needed for https://github.com/oat-sa/extension-tao-scheduler/pull/41


common/ext/class.Extension.php
```diff
@@ -493,4 +493,23 @@ public function getConfigHeader($key)
         }
         return null;
     }
+
+    /**
+     * @return common_ext_ExtensionUpdater
+     * @throws common_ext_ManifestException
+     * @throws common_ext_ManifestNotFoundException
+     */
+    public function getUpdater()
```
jbout, line 502:
```suggestion
     * @throws common_ext_ManifestException
     * @throws common_ext_ManifestNotFoundException
     */
    public function getUpdater(): common_ext_ExtensionUpdater
```

PR #799, opened 2020-06-11
Fix/multiple extension manager

jbout:
Fixes issue during install where the cached version in common_ext_ExtensionsManager does not correspond to the extension manager in service locator

### DIFF
--- a/common/ext/class.ExtensionsManager.php
+++ b/common/ext/class.ExtensionsManager.php
@@ -55,14 +55,6 @@ class common_ext_ExtensionsManager extends ConfigurableService
     private $extensions = [];
 
     /**
-     * Singleton instance of common_ext_ExtensionsManager
-     *
-     * @access private
-     * @var common_ext_ExtensionsManager
-     */
-    private static $instance = null;
-
-    /**
      * @deprecated Use ServiceManager::get(\common_ext_ExtensionsManager::SERVICE_ID) instead
      *
      * Obtain a reference on a unique common_ext_ExtensionsManager
@@ -74,15 +66,7 @@ class common_ext_ExtensionsManager extends ConfigurableService
      */
     public static function singleton()
     {
-        if (! isset(self::$instance)) {
-            if (ServiceManager::getServiceManager()->has(self::SERVICE_ID)) {
-                self::$instance = ServiceManager::getServiceManager()->get(self::SERVICE_ID);
-            } else {
-                self::$instance = new common_ext_ExtensionsManager();
-                ServiceManager::getServiceManager()->propagate(self::$instance);
-            }
-        }
-        return self::$instance;
+        return ServiceManager::getServiceManager()->get(self::class);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


common/ext/class.ExtensionsManager.php
```diff
@@ -74,15 +66,7 @@ class common_ext_ExtensionsManager extends ConfigurableService
...
-        if (! isset(self::$instance)) {
-            if (ServiceManager::getServiceManager()->has(self::SERVICE_ID)) {
-                self::$instance = ServiceManager::getServiceManager()->get(self::SERVICE_ID);
-            } else {
-                self::$instance = new common_ext_ExtensionsManager();
-                ServiceManager::getServiceManager()->propagate(self::$instance);
-            }
-        }
-        return self::$instance;
+        return ServiceManager::getServiceManager()->get(self::class);
```
peetya, line 69: How can you ensure that the `ExtensionManager` is already exist in the `ServiceManager`?

jbout: It is the first service we register:
https://github.com/oat-sa/generis/blob/master/common/oatbox/install/Installer.php#L51

core/kernel/classes/class.Container.php
```diff
@@ -26,21 +26,7 @@
...
-
-
-    /**
-     * Short description of method __serialize
-     *
-     * @access public
-     * @author firstname and lastname of author, <author@example.org>
-     * @return string
-     */
-    public function __serialize()
```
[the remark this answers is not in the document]
jbout: The function is causing a php error as its signature does not match
https://www.php.net/manual/en/language.oop5.magic.php#object.serialize
hence breaking the unit test. If it were used anywhere else it would break that code too.

So while in general I agree that changes to public functions represent backward breaking changes, in this case I actually restored the functionality: https://github.com/oat-sa/generis/pull/799/files#diff-19cb0be98baad0241be1c7fc43f65ea7R770

test/GenerisTestCase.php
```diff
@@ -22,25 +22,18 @@
...
-            'testfs' => class_exists('League\Flysystem\Memory\MemoryAdapter')
-                ? [
-                    'class' => \League\Flysystem\Memory\MemoryAdapter::class
-                ]
-                : [
-                    'class' => FileSystemService::FLYSYSTEM_LOCAL_ADAPTER,
-                    'options' => ['root' => \tao_helpers_File::createTempDir()]
-                ]
+            'testfs' => [
+                'class' => MemoryAdapter::class
```
hutnikau, line 35: Since we have memory adapter as a dependency of generis we could remove check that this class exists in other places as well:
https://github.com/oat-sa/generis/blob/master/test/integration/common/filesystem/DirectoryFilesystemTest.php#L100
https://github.com/oat-sa/generis/blob/master/test/integration/helpers/FileSerializerMigrationHelperTest.php#L180

jbout: done

PR #800, opened 2020-06-11
[UNO-329] Feature: added a TTL option to KeyValueSessionHandler

SergiiTao:
This improvement is implemented in order to make session TTL configurable from a client extension without touching php.ini


jbout:
Will this not result in different session times depending on the implementation, as this will only have an effect if the keyvalue session is used?

SergiiTao:
> Will this not result in different session times depending on the implementation, as this will only have an effect if the keyvalue session is used?

as we can have only one session handler at the same time and default behavior remains the same I don't expect any issues with that.

PR #803, opened 2020-06-16
Feature UNO-339 add search text box definition to the widget RDF representation

wazelin:
[UNO-339](https://oat-sa.atlassian.net/browse/UNO-339) feat: introduce `http://www.tao.lu/datatypes/WidgetDefinitions.rdf#SearchTextBox` widget's RDF definition
[UNO-339](https://oat-sa.atlassian.net/browse/UNO-339) fix(dependencies): stop using a `tao` class in tests

This PR doesn't really do anything on its own, it's purpose is to introduce a new widget entry into the RDF structure.

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -39,6 +39,7 @@ use oat\generis\model\kernel\persistence\smoothsql\search\ComplexSearchService;
 use oat\generis\model\kernel\uri\UriProvider;
 use oat\generis\model\user\AuthAdapter;
 use oat\generis\model\user\UserFactoryService;
+use oat\generis\model\WidgetRdf;
 use oat\generis\persistence\PersistenceManager;
 use oat\oatbox\action\ActionService;
 use oat\oatbox\cache\KeyValueCache;
@@ -528,9 +529,10 @@ class Updater extends common_ext_ExtensionUpdater
         $this->skip('12.23.0', '12.26.1');
 
         if ($this->isVersion('12.26.1')) {
-            $widgetDefinitionsFilePath =  __DIR__ . '/../../core/ontology/widgetdefinitions.rdf';
-            $api = core_kernel_impl_ApiModelOO::singleton();
-            $api->importXmlRdf('http://www.tao.lu/datatypes/WidgetDefinitions.rdf', $widgetDefinitionsFilePath);
+            core_kernel_impl_ApiModelOO::singleton()->importXmlRdf(
+                WidgetRdf::NAMESPACE,
+                __DIR__ . '/../../core/ontology/widgetdefinitions.rdf'
+            );
 
             $this->setVersion('12.27.0');
         }
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -526,6 +526,18 @@ class Updater extends common_ext_ExtensionUpdater
         }
 
         $this->skip('12.23.0', '12.26.1');
+
+        if ($this->isVersion('12.26.1')) {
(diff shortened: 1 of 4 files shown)


scripts/update/Updater.php
```diff
@@ -527,6 +527,17 @@ public function update($initialVersion)
 
         $this->skip('12.23.0', '12.24.1');
 
+        if ($this->isVersion('12.24.1')) {
+            $file = __DIR__ . DIRECTORY_SEPARATOR .
```
Alroniks, line 531: Why we should use DIRECTORY_SEPARATOR here? It may sense to use it for extract part from a real path on Windows, but for composing Linux-like separator ("/") will work well on both platforms.

Another approach, using a straight path with __DIR__ '/path/to/file' allows quickly jump to a real file in the IDE and see highlights when typing the path.

scripts/update/Updater.php
```diff
@@ -527,6 +527,17 @@ public function update($initialVersion)
...
+                '..' . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR .
+                'core' . DIRECTORY_SEPARATOR .
+                'ontology' . DIRECTORY_SEPARATOR .
+                'widgetdefinitions.rdf';
```
Alroniks, line 535:
```suggestion
            $widgetDefinitionsFilePath =  __DIR__ . '/../../core/ontology/widgetdefinitions.rdf';
```

scripts/update/Updater.php
```diff
@@ -527,6 +527,17 @@ public function update($initialVersion)
...
+            $api = core_kernel_impl_ApiModelOO::singleton();
+            $api->importXmlRdf('http://www.tao.lu/datatypes/WidgetDefinitions.rdf', $file);
```
Alroniks, line 537:
```suggestion
            $api->importXmlRdf('http://www.tao.lu/datatypes/WidgetDefinitions.rdf', $widgetDefinitionsFilePath);
```

scripts/update/Updater.php
wazelin: We shouldn't. It's just a slightly modified copy of a similar block up above ;)

scripts/update/Updater.php
```diff
@@ -526,6 +526,14 @@ public function update($initialVersion)
         }
 
         $this->skip('12.23.0', '12.25.0');
+
+        if ($this->isVersion('12.25.0')) {
+            $widgetDefinitionsFilePath =  __DIR__ . '/../../core/ontology/widgetdefinitions.rdf';
+            $api = core_kernel_impl_ApiModelOO::singleton();
+            $api->importXmlRdf('http://www.tao.lu/datatypes/WidgetDefinitions.rdf', $widgetDefinitionsFilePath);
```
Alroniks, line 533:
```suggestion
            $core_kernel_impl_ApiModelOO::singleton()->importXmlRdf(
                WidgetRdf::NAMESPACE, 
                __DIR__ . '/../../core/ontology/widgetdefinitions.rdf'
            );
```

Sorry for annoying, but it seems we can avoid usage variables at all and improve a bit WidgetRdf interface, adding the new constant NAMESPACE and having something like that:

```
interface WidgetRdf
{
    public const NAMESPACE = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf';
    public const CLASS_URI_WIDGET = self::NAMESPACE . '#WidgetClass';
//    ...more code here
}
```

wazelin: No annoyance caused at all, your suggestion makes perfect sense, thank you!

PR #820, opened 2020-07-13
Feature/nex 1135/sql iterators types support

boajer:
https://oat-sa.atlassian.net/browse/NEX-1135

* Allows specifying limits
* Allows more flexible parametrization by explicit param types specification  ( compatibility with PDO improved ) 
* get rid of redundent PHPDoc

### DIFF
--- a/common/persistence/sql/class.QueryIterator.php
+++ b/common/persistence/sql/class.QueryIterator.php
@@ -31,9 +31,13 @@ class common_persistence_sql_QueryIterator implements Iterator
      * @var common_persistence_SqlPersistence
      */
     private $persistence;
-
+    /**
+     * @var string
+     */
     private $query;
-
+    /**
+     * @var array
+     */
     private $params;
 
     /**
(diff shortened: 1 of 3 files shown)


siwane:
Nice cleaning 👍

common/persistence/sql/class.QueryIterator.php
```diff
@@ -14,99 +14,78 @@
...
     /**
      * @var common_persistence_SqlPersistence
      */
     private $persistence;
-    
-    /**
-     * Query to iterator over
-     *
-     * @var string
-     */
```
bartlomiejmarszal, line 40: Why did you remove them?

common/persistence/sql/class.QueryIterator.php
```diff
@@ -14,99 +14,78 @@
...
-    public function __construct(common_persistence_SqlPersistence $persistence, $query, $params = [])
-    {
+    private $cache;
+    private $types;
+    private $limit;
+
+    public function __construct(
+        common_persistence_SqlPersistence $persistence,
+        $query,
+        $params = [],
```
gabrielfs7, line 58:
```suggestion
        string $query,
        array $params = [],
```

Now we can add type hint :)

common/persistence/sql/class.QueryIterator.php
gabrielfs7: @boajer  As we do not have php7.4 we still need `@var` php doc for attributes. When you put it back, I would remove the useless comments and fix the `$params` which is an array, not a string.

PR #821, opened 2020-07-15
TAO-10365 Added new interface for retrieving access data from resources

luisfn:
Added a new interface to help distinguish when we have the possibility to getResourceAccessData from the PermissionProvider

### DIFF
--- a/core/data/permission/ReverseRightLookupInterface.php
+++ b/core/data/permission/ReverseRightLookupInterface.php
@@ -32,7 +32,21 @@ use core_kernel_classes_Resource;
 interface ReverseRightLookupInterface
 {
     /**
-     * Returns a list roles related to a resource
+     * Returns a list roles and permissions related to a resource
+     *
+     *
+     * Sample data to be returned:
+     * [
+     *    'http://www.tao.lu/Ontologies/TAO.rdf#BackOfficeRole' => [
+     *        'GRANT',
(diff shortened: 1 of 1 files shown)


core/data/permission/ReverseRightLookupInterface.php
```diff
@@ -0,0 +1,53 @@
...
+     *    'http://www.tao.lu/Ontologies/TAO.rdf#BackOfficeRole' => [
+     *        'GRANT',
+     *        'READ',
+     *        'WRITE'
+     *    ],
+     *    'http://www.tao.lu/Ontologies/TAO.rdf#ItemAuthor' => [
+     *        'GRANT',
+     *        'READ'
+     *    ],
+     * ]
```
wazelin, line 47: It's better to extract those into a `PermissionCollection` entity.

luisfn: > Don't forget the `manifest.php`.

It won't be done right now but you are right, we will check it when merging `feature/TAO-10203-advanced-search` to `develop` 

> It's better to extract those into a `PermissionCollection` entity.
> Without it, the interface doesn't really enforce any contract :)

I totally agree, kind o hate these arrays everywhere but unfortunately it's being used in many places, so I can't change it right now (not the focus here)

PR #824, opened 2020-07-29
Labels of triples with UI translates

alroniks:
The related task is [BSA-145](https://oat-sa.atlassian.net/browse/BSA-145)

It removes deprecated `label` property as part of refactoring and brings the ability to get translated labels if UI translation exists.

### DIFF
--- a/core/kernel/classes/class.Resource.php
+++ b/core/kernel/classes/class.Resource.php
@@ -62,17 +62,6 @@ class core_kernel_classes_Resource extends core_kernel_classes_Container
     public $uriResource = '';
 
     /**
-     * The resource label
-     * direct access to the label is deprecated,
-     * please use getLabel()
-     *
-     * @access public
-     * @var string
-     * @deprecated
-     */
-    public $label = null;
-
-    /**
      * The resource comment
      * direct access to the comment is deprecated,
      * please use getComment()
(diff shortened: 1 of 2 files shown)


core/kernel/classes/class.Resource.php
```diff
@@ -187,40 +176,32 @@ public function getTypes()
...
-        if (is_null($this->label)) {
-            $label =  $this->getOnePropertyValue($this->getProperty(OntologyRdfs::RDFS_LABEL));
-            $this->label = is_null($label)
-                ? ''
-                : ($label instanceof core_kernel_classes_Resource
-                    ? $label->getUri()
-                    : (string)$label
-                   )
-            ;
+    public function getLabel(): ?string
```
wazelin, line 183: I don't see a way of getting a `null` as a return value here.

```suggestion
    public function getLabel(): string
```

core/kernel/classes/class.Resource.php
```diff
@@ -61,17 +61,6 @@ class core_kernel_classes_Resource extends core_kernel_classes_Container
...
     public $uriResource = '';
 
-    /**
-     * The resource label
-     * direct access to the label is deprecated,
-     * please use getLabel()
-     *
-     * @access public
-     * @var string
-     * @deprecated
```
wazelin, line 71: Even though the property usage has been deprecated, that modification is a breaking change.

manifest.php
```diff
@@ -32,7 +32,7 @@
     'label' => 'Generis Core',
     'description' => 'Core extension, provide the low level framework and an API to manage ontologies',
     'license' => 'GPL-2.0',
-    'version' => '13.6.0',
+    'version' => '13.7.0',
```
wazelin, line 35: Due to a breaking change, mentioned above, please consider bumping the major version.

```suggestion
    'version' => '14.0.0',
```

wazelin (changes requested):

core/kernel/classes/class.Resource.php
```diff
@@ -187,40 +176,32 @@ public function getTypes()
...
+    {
+        $label =  $this->getOnePropertyValue($this->getProperty(OntologyRdfs::RDFS_LABEL));
+
+        if ($label === null) {
+            return '';
         }
-        
-        return $this->label;
+
+        return $label instanceof self ? $label->getUri() : __((string)$label);
```
SergiiTao, line 191: I am afraid that it can break some translations like `__('%s Class saved', $clazz->getLabel())` cause label can be translated and we don't have a translation for half translated text

also we need to care of cases like `__\(\$.*->getLabel` for example `__($property->getLabel())`

SergiiTao (changes requested):

core/kernel/classes/class.Resource.php
Alroniks: As an option, I can add a parameter to `getLabel` with the default option, and if need to turn off translation, the code user can configure it during the call. Something like this:
```
public function getLabel(bool $translate = true) {}
// ...
$v = __($property->getLabel(false));
```

Alroniks: As it already breaks BC, adding a new property with default option seems pretty safe.

wazelin (approved):
- [x] Feature is working correctly on my local machine (if applicable)

core/kernel/classes/class.Resource.php
SergiiTao: default `true` will do the same. it will break a lot of translations.
but default `false` is not better than __($property->getLabel()). you still need to go through all extension and put this `true` 
anyway, I believe this is not a responsibility of the Resource class to translate labels complexity and coupling getting higher.

ktsirkou: Translating the label should be restricted on TAO internal resources, not on all resources. 
Example: define a new class with Label = Label or any other keyword we translate. It will get translated as well. We need to  prohibit this behavior.

jbout: We already have a system in place to translate labels, please check how the column "l_language" is used, and this can work on all resources, even user defined resources.

Please do not use the code translation tool on top of the ontology translation, this could have very weird effects

jbout (changes requested):
Please use generis functionality such as "getPropertyValuesByLg" to translate labels. **__()** should never be used with variable strings.

core/kernel/classes/class.Resource.php
jbout: Also @SergiiTao this is not how __('%s Class saved', $clazz->getLabel()) works. The translation is applied before the string replacement.

Users are asked to translate '%s Class saved' so it will work with any class label. Please keep using this format and never use construct such as:
`__($clazz->getLabel() . ' Class saved')`
or 
`$clazz->getLabel() . __(' Class saved')`

core/kernel/classes/class.Resource.php
jbout: has the performance impact been considered by removing the label cache?

PR #826, opened 2020-08-20
Fix/tao 10423 label not indexed in memory queue

andreluizmachado:
This PR is intended to fix the issues:
https://oat-sa.atlassian.net/browse/TAO-10423
https://oat-sa.atlassian.net/browse/TAO-10318

**The problem**
Tao platforms are sending duplicate ResourceUpdate events. This is because, in most cases, resources are not changed in a transaction, but in small blocks. The result of this is a storm of duplicate events.

**The solution**
We will aggregate only single events and trigger them when we reach the maximum number of aggregated events or when the application is shutting down (check: https://github.com/oat-sa/tao-core/pull/2646 for more details).


core/kernel/classes/class.Resource.php
```diff
@@ -704,11 +706,9 @@ public function getUri()
      * @param  Resource resource
      * @return boolean
      */
-    public function equals(core_kernel_classes_Resource $resource)
+    public function equals(core_kernel_classes_Resource $resource): bool
     {
-        $returnValue = (bool) false;
-        $returnValue = $this->getUri() == $resource->getUri();
-        return (bool) $returnValue;
+        return $this->getUri() == $resource->getUri();
```
luisfn, line 711: Maybe use `===`?

manifest.php
```diff
@@ -32,7 +32,7 @@
     'label' => 'Generis Core',
     'description' => 'Core extension, provide the low level framework and an API to manage ontologies',
     'license' => 'GPL-2.0',
-    'version' => '13.1.0',
+    'version' => '13.1.1',
```
luisfn, line 35: Until now we were not updating versions and it probably will be outdated until merge to develop/master so I'm not sure we should update now

migrations/Version202008201138552348_generis.php
```diff
@@ -0,0 +1,39 @@
...
+    public function getDescription(): string
+    {
+        return 'New EventAggregator Service for extension generis.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $eventAggregator = new EventAggregator(
+          [
+              'numberOfAggregatedEvents' => 10
```
luisfn, line 25: Maybe call it maxAggregatedEvent? the current name is fine as well, if you think it's better just keep it

test/OntologyMockTrait.php
```diff
@@ -83,8 +86,10 @@ private function setupOntology(Ontology $onto)
             EventManager::SERVICE_ID => new EventManager(),
             LoggerService::SERVICE_ID => $this->prophesize(LoggerInterface::class)->reveal(),
             UriProvider::SERVICE_ID => new Bin2HexUriProvider([Bin2HexUriProvider::OPTION_NAMESPACE => 'http://ontology.mock/bin2hex#']),
-            SimpleCache::SERVICE_ID => new NoCache()
+            SimpleCache::SERVICE_ID => new NoCache(),
+            EventAggregator::SERVICE_ID => $eventAggregator
         ]);
+        $eventAggregator->setServiceLocator($sl);
```
luisfn, line 92: I know the name was already there, but would rename `$sl` and `$pm` to meaningful names

test/unit/oatbox/event/EventAggregatorTest.php
```diff
@@ -0,0 +1,114 @@
...
+
+        $this->serviceLocator = $this->getServiceLocatorMock(
+            [
+                EventManager::SERVICE_ID => $this->eventManager,
+                LoggerService::SERVICE_ID => $this->logger,
+            ]
+        );
+    }
+
+    public function testTriggerEventsWhenNumberOfAggregatedEventsRichesTheMaximum(): void
```
luisfn, line 58: Reaches

test/unit/oatbox/event/EventAggregatorTest.php
```diff
@@ -0,0 +1,114 @@
...
+            ->method('trigger');
+
+        $this->logger->expects($this->never())->method('info');
+
+        $eventAggregator = new EventAggregator(['numberOfAggregatedEvents'=>4]);
+        $eventAggregator->setServiceLocator($this->serviceLocator);
+        $eventAggregator->triggerAggregatedEvents();
+    }
+
+    public function testTriggerAggregatedEvents(): void
```
luisfn, line 95: Here you are triggering even if the maximum was not reached? Like forcing?
If yes, maybe you should rename the test

luisfn (changes requested):
Seems pretty good to me, would just expect a review from someone with more experience on tao
Aside from that, have added just a couple questions

common/oatbox/event/EventAggregator.php
```diff
@@ -0,0 +1,79 @@
...
+    private $events = [];
+
+    public function __construct(array $options = [])
+    {
+        parent::__construct($options);
+
+        $this->numberOfAggregatedEvents = $options['numberOfAggregatedEvents'];
+    }
+
+    public function put(string $eventId, Event $event): void
```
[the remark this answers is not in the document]
andreluizmachado: I thought about it and the problem is that would impact all the events which is a big risk IMO.

Skyld (approved):
- [x] New code is covered by tests
- [ ] Tests are running successfully (old and new ones) on my local machine
- [x] New code is not subject to concurrency issues
- [ ] Feature is working correctly on my local machine
- [ ] Changelog is updated according to changes
- [ ] Documentation is updated according to changes

PR #828, opened 2020-08-26
Feature/tao 10287 Fire ResourceUpdated Event When Class type changes.

andreluizmachado:
This implementation is intended to cover the ticket https://oat-sa.atlassian.net/browse/TAO-10287.

**Before this implementation:**
- ResourceUpdated Event is not fired when the type of the resource is removed or set
- Some Methods are not well-formatted and with no return type hints

**After this implementation:**
- ResourceUpdated Event is now fired when the type of the resource is removed or set. This reflects better the goal of the event and it'll automatically reindex the resources on advanced search when a resource is moved from a class to another one.
 - Some Methods are now well-formatted and with return type hints


Skyld (approved):
- [ ] New code is covered by tests
- [ ] Tests are running successfully (old and new ones) on my local machine
- [x] New code is not subject to concurrency issues
- [ ] Feature is working correctly on my local machine
- [ ] Changelog is updated according to changes
- [ ] Documentation is updated according to changes

luisfn (approved):

PR #831, opened 2020-08-28
Lib dependency version bumped

boajer:


composer.json
```diff
@@ -63,7 +63,7 @@
     "league/flysystem": "~1.0",
     "league/flysystem-memory": "~1.0",
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
-    "oat-sa/lib-generis-search": "2.1.2",
+    "oat-sa/lib-generis-search": "2.2.0",
```
gabrielfs7, line 66: I think we should update it later when we are sure this spanner fix will not break the search.

mike-ionut-mihai-sandu-tao: it is ok, the only change was about the phpunit upgrade

PR #836, opened 2020-09-10
Isolated unit test configuration

gitromba:
Recently we have started using the `config/sample/generis.conf.php` to "mock" some essential global constants and so to overcome some issues which are causing false negative broken unit tests.

It turned out that this sample file, despite its name, being used during the TAO install. This tight coupling with [tao-core](https://github.com/oat-sa/tao-core/blob/cadb4fb40b52b3a648d9c28f2c39c55f8c689bd8/install/class.Installator.php#L186) prevents one to change configurations without risking side effects in the installation process.

This PR intends to bring a solution by introducing an isolated configuration specifically used for tests. Essentially it's the copy of `config/sample/generis.conf.php` and the effects of updating this file is isolated to tests only.


test/bootstrap.php
```diff
@@ -26,4 +26,4 @@
 // Use TAO dependency resolver first.
 require_once $extensionRoot . './../vendor/autoload.php';
 
-common_Config::load($extensionRoot . 'config/sample/generis.conf.php');
+common_Config::load($extensionRoot . 'test/generis.conf.php');
```
gabrielfs7, line 29:
```suggestion
common_Config::load($extensionRoot . 'test/config/generis.conf.php');
```
Wdyt?

gitromba: Good point, makes sense!

I'll update it shortly.

test/generis.conf.php
```diff
@@ -0,0 +1,81 @@
...
+
+# local namespace
+define('LOCAL_NAMESPACE', '');
+
+# platform identification
+define('GENERIS_INSTANCE_NAME', '');
+define('GENERIS_SESSION_NAME', '');
+
+# paths
+define('ROOT_PATH', '');
```
gabrielfs7, line 42: I think the ROOT_PATH could be predefined as `__DIR__  . DIRECTORY_SEPARATOR .  '..' . DIRECTORY_SEPARATOR .  '..' . DIRECTORY_SEPARATOR`;

I am afraid some test would be loading something from this path which will no longer exit anymore. I am not sure about this too. Wdyt?

gabrielfs7 (approved):
LGTM

test/generis.conf.php
gitromba: It would be very wrong if any unit test used this constant. So far I haven't seen any sign of it luckily. 

It just occurred to me now: what if I remove (comment) those ones which we know for sure doesn't used. To embrace the vision, that is for legacy only and for new tests proper DI must be in place.

WDYT?

gitromba: It would be very wrong if any unit test did depend on this constant. Luckily I haven't found any.

It just occurred to me: Is it better if I just remove all the "placeholder" constants (with no proper value) to embrace our vision that this is for legacy only?

WDYT?

mike-ionut-mihai-sandu-tao (approved):

test/generis.conf.php
gabrielfs7: Yes, that is why I was not sure what I suggested. Thanks for the reply. I also think I cannot calculate right know if it will break or not some of the extensions we have. I think with time we will discover. It could be empty then.

gitromba: @gabrielfs7 thanks for sharing your thoughts. 

WDYT on having all the "empty" constants removed from this config? For instance `define('DEBUG_MODE', false);` would stay as it has proper value set, but `define('ROOT_URL', '');` would be gone as it has an obviously invalid (empty) value. Same filter to be applied on all constants. At then end we would end up having a few lines long config containing only relevant entries which is in addition also aligned with our "vision".

PR #843, opened 2020-09-30
Token handling on GCP 

gabrielfs7:
Provide support to single cache option to manage auth token on GCP:

We must use proper key `GCP-TOKEN-SANCTUARY:GOOGLE_AUTH_PHP_GCE`

```
xvzf on gke_tao-staging-281916_europe-west1_staging-gke (udir-qa) in ~ ❯ k exec -it deliver-be-697cd57f7f-x449w -c redis -- redis-cli
127.0.0.1:6379> keys *
1) "GCP-TOKEN-SANCTUARY:GOOGLE_AUTH_PHP_GCE"
127.0.0.1:6379> GET GCP-TOKEN-SANCTUARY:GOOGLE_AUTH_PHP_GCE
"a:3:{s:12:\"access_token\";s:454:\"XXXXXXXXXXXXXXXXXXX\";s:10:\"expires_in\";i:3599;s:10:\"token_type\";s:6:\"Bearer\";}"
127.0.0.1:6379>
```

IMPORTANT:

- Must use code on this branch to test: https://github.com/oat-sa/nextgen-backoffice/pull/5
- Necessary to clear Redis cache in order to run migrations and test it.
- Remove application cache `rm -rf data/generis/cache/*`
- Enable writing on 
```
<?php
return new oat\oatbox\cache\GcpTokenCacheItemPool(array(
    'persistence' => 'gcpTokenKeyValue',
    'enableDebug' => true,
    'disableWrite' => false, //@TODO This must be a false for develop
    'tokenCacheKey' => 'GCP-TOKEN-SANCTUARY:GOOGLE_AUTH_PHP_GCE'
));
```

https://oat-sa.atlassian.net/browse/NEX-2058


common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,251 @@
...
+use oat\oatbox\service\ConfigurableService;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+class GcpTokenCacheItemPool extends ConfigurableService implements CacheItemPoolInterface
+{
+    public const SERVICE_ID = 'generis/GcpTokenCacheItemPool';
+
+    public const OPTION_PERSISTENCE = 'persistence';
+    public const OPTION_ENABLE_DEBUG = 'enableDebug';
```
siwane, line 35: Today we already have a constant `DEBUG_MODE` for this purpose. I agree that it is not the ideal, but we should not duplicate it. We should have a centralized place to switch the platform on debug or not

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,251 @@
...
+    private function serializeToPersist(CacheItemInterface $item): void
+    {
+        if (is_array($item->get())) {
+            $item->set(serialize($item->get()));
+        }
+    }
+
+    private function getCacheKey(): string
+    {
+        return (string)$this->getOption(self::OPTION_TOKEN_CACHE_KEY, 'GCP-TOKEN-SANCTUARY:GOOGLE_AUTH_PHP_GCE');
```
siwane, line 244: Should we put this value into constant?

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,251 @@
...
+        );
+    }
+
+    private function unSerializeIfNecessary(CacheItemInterface $item): void
+    {
+        if (empty($item->get())) {
+            return;
+        }
+
+        $unSerialized = @unserialize($item->get());
```
siwane, line 228: I have some fear about this method. In the case of some snippets are caching serialized data, it will be unserialized here and a second time in the usage of this cache. WDYT?

common/oatbox/cache/ItemPoolSimpleCacheAdapter.php
```diff
@@ -132,6 +134,10 @@ public function commit()
 
     private function getCache(): CacheInterface
     {
+        if ($this->hasOption(self::OPTION_CACHE_SERVICE)) {
```
siwane, line 137: What do u think about putting this GCPCache in default configuration (in place of SimpleCache::SERVICE_ID) instead of hijacking the dependency here?
It will be set only on GCP environment ofc

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,251 @@
...
+        if ($this->isWritingDisabled === null) {
+            $this->isWritingDisabled = (bool)$this->getOption(self::OPTION_DISABLE_WRITE, false);
+        }
+
+        return $this->isWritingDisabled;
+    }
+
+    private function getCache(): CacheItemPoolInterface
+    {
+        $this->cache = $this->createCache();
```
siwane, line 208: Do u think we should check if this attribute has a value before to avoid to recreate it if it already exists?

common/persistence/DriverConfigurationFeeder.php
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ */
+
+declare(strict_types=1);
+
+namespace oat\generis\persistence;
+
+use oat\oatbox\service\ConfigurableService;
+
+class DriverConfigurationFeeder extends ConfigurableService
```
siwane, line 27: I do not get the value of this class, could we just declare a spannerDriver in the persistence.conf.php. As far as I remember it is the purpose of this library driver, isn't it?

siwane (changes requested):

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: Hmm. I will not use this global constant here. I could use at the `conf.php`, but still is a global constant what we want to avoid. This config is based on DI though. This is not code duplication.

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: I could, but would be only used here. Do you think it worth?

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: This cache is used only for GCP token, not other code. I would prefer json_encoded, but It is not possible right now. Do you have any suggestion that would work with this?

common/oatbox/cache/ItemPoolSimpleCacheAdapter.php
gabrielfs7: You mean convert `ItemPoolSimpleCacheAdapter.php` to a configurable service, create a `.conf.php` and add it in the migrations / installation?

What is the difference between this? It is not hijacking. This is the tao DI, I can set options when I build service. Did not see the problem doing it.

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: Good point, I will had this and removed after refactoring :)

common/persistence/DriverConfigurationFeeder.php
gabrielfs7: Can you show me how you would to it with the needs I have here? I did not find any way of doing that.

common/oatbox/cache/GcpTokenCacheItemPool.php
siwane: I understand and I would not recommend to use the global constant. But if someone is changing to true or false, it will not have any effect on your service.

gabrielfs7: I added in the conf.php file, but if you do not mind will not add this to the class cause we are making an effort to remove global constants from our codebase.

boajer: Instead of constant `DEBUG_MODE`,  `\oat\tao\model\service\ApplicationService::isDebugMode` may be used

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,253 @@
...
+            return true;
+        }
+
+        return $this->getCache()->commit();
+    }
+
+    private function log(string $method, array $keys = []): void
+    {
+        if ($this->isDebug()) {
+            $this->getLogger()->debug(
```
boajer, line 177: What is the purpose? If logger has been configured to keep `debug` level - we are already fine; 
Most of the time production envs will not log anything, even if this particular service is in debug mode, as logger verbosity used to at about ~ `warn` level

common/persistence/DriverConfigurationFeeder.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    private const DRIVER_OPTIONS_REPLACE = [
+        'OAT\Library\DBALSpanner\SpannerDriver' => [
```
boajer, line 30: Would you consider `SpannerDriver::class` instead of  hardcoded string?

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: This is a good point, indeed we do not need this if. I will remove! Thanks for the heads up!

common/persistence/DriverConfigurationFeeder.php
gabrielfs7: Would not do this, cause might be the case the lib is not installed.

common/oatbox/cache/GcpTokenCacheItemPool.php
gabrielfs7: Closing, cause I removed!

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,238 @@
...
+
+    /** @var bool */
+    private $isWritingDisabled;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItem($key)
+    {
+        $key = $this->getCacheKey();
```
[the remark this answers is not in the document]
bartlomiejmarszal: Why not keep parameter from method signature and instead key would be passed from configurable service where we can get GCP key that we want to retrieve.

common/persistence/DriverConfigurationFeeder.php
```diff
@@ -0,0 +1,72 @@
...
+            'driverOptions' => [
+                'driver-option-auth-pool',
+                'driver-option-session-pool'
+            ]
+        ]
+    ];
+
+    public function feed(array $config): array
+    {
+        if (empty($config['connection']['driverClass'])) {
```
bartlomiejmarszal, line 40: Why we are checking if this is empty and then if it is we return it? We want to ensure that our configuration doesn't not have this value?

common/persistence/DriverConfigurationFeeder.php
```diff
@@ -0,0 +1,72 @@
...
+    private const DRIVER_OPTIONS_REPLACE = [
+        'OAT\Library\DBALSpanner\SpannerDriver' => [
+            'driverOptions' => [
+                'driver-option-auth-pool',
+                'driver-option-session-pool'
+            ]
+        ]
+    ];
+
+    public function feed(array $config): array
```
bartlomiejmarszal, line 38: Is the purpose of this method to replace some part of configuration?

common/persistence/DriverConfigurationFeeder.php
bartlomiejmarszal: Why don't we feed this config with service values?

bartlomiejmarszal (commented):
- [x] New code is covered by tests (if applicable)

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,238 @@
...
+    {
+        $this->log(__METHOD__);
+
+        return $this->getCache()->clear();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItem($key)
```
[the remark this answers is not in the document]
gabrielfs7: No, for delete we still need. This approach is used by deliver team on production. I am not inventing it.

common/persistence/DriverConfigurationFeeder.php
gabrielfs7: If you give me a working example I can do it, but tao does not support it for drivers.

common/persistence/DriverConfigurationFeeder.php
gabrielfs7: We cannot inject it dynamically for Drivers, cause they are not configurable services and it is based on class name.

common/oatbox/cache/GcpTokenCacheItemPool.php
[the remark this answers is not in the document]
bartlomiejmarszal: We had a conversation with @gabrielfs7 where he exaplin me why we are doing this like that. 
The purpose of this hardcoding is to isolate GcpTokenCacheItemPool from retrieving any other key then configured in options. This key is set in Sanctuary and can not be changed. 

I'm resolving this comment

common/oatbox/cache/GcpTokenCacheItemPool.php
siwane: If its only used by GCP, the risk is low 👍

config/default/GcpTokenCacheItemPool.conf.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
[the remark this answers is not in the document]
siwane: true, it is why I think we should install it only when needed, via seed file and only on spanner env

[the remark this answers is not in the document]
gabrielfs7: What do you suggest? To delete the `.conf.php` file? I think we will have bugs in the platform. I would like to avoid another 1h installation to check it. Even if this file is there it will not be used, so I do not see a risk. Also, when we have a class with `SERVICE_ID` we by standard create a conf.php. It would be strange in this case to not do it.

@mike-ionut-mihai-sandu-tao @bartlomiejmarszal @mike-ionut-mihai-sandu-tao Can you guys give some input on this, please?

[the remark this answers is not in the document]
siwane: For example, on AWS environment, and only on this kind of env, we have `awsClient.conf.php`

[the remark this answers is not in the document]
mike-ionut-mihai-sandu-tao: if we can leave out the initial default config and put this inside the readme then ok, but overall some registration of that information will be there anyway ,  so i don't consider this an impediment to close the pr
- if we strictly refer to the fact that we are adding 1 more class and that this will increase maintenance i agree
- if we refer that usually the learning curve  is a bit bigger and is hard to find configs and the fact that documentation is not proper i agree also
so from my point of view : 
 - let's leave the default config or make 100% sure that this is easy to configure and that the information is reachable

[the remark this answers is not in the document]
gabrielfs7: For me it is just a config there that will only be used if we call the SpannerDriver, otherwise should not cause any issue. The conf.php is basically our "DI" mapping. You do not remove things from DI mapping just because you are not using it. So IMO we should maintain.

common/oatbox/cache/ItemPoolSimpleCacheAdapter.php
mike-ionut-mihai-sandu-tao: i don't see any benefit of doing the change

common/persistence/DriverConfigurationFeeder.php
mike-ionut-mihai-sandu-tao: if you put the class you create a mandatory requirement of that library/ class and we don't want that

mike-ionut-mihai-sandu-tao (approved):
lgtm

config/default/GcpTokenCacheItemPool.conf.php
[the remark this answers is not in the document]
gabrielfs7: Discussed in person with @siwane . I removed the migrations file with this.

common/oatbox/cache/GcpTokenCacheItemPool.php
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        return $this->cache;
+    }
+
+    private function createCache(): CacheItemPoolInterface
+    {
+        return $this->getCacheFactory()->create(
+            [
+                CacheItemPoolFactory::CONFIG_PERSISTENCE => $this->getOption(self::OPTION_PERSISTENCE, 'redis'),
```
boajer, line 204: I have doubts if we always have ` 'redis'` persistence and redis in general at the application stack.
AFAIK we have `default_kv` as a representation

gabrielfs7: I agree, thanks!

siwane (approved):
- [x] New code is covered by tests (if applicable)

boajer (approved):
- [x] New code is covered by tests (if applicable)

PR #848, opened 2020-10-12
Fix/cache sessions

tikhanovichA:
Because of https://github.com/oat-sa/generis/pull/791/files#diff-41527d5edb3851cc22277782cd23b829R68 when update the `config/generis/SimpleCache.conf.php` and set `persistence = redis` I see the next error message `Includes roles related to Role with URI '${roleUri}' is not in the Cache memory.`


jbout (approved):
good catch!

hutnikau (approved):

PR #852, opened 2020-10-20
Remove event aggregator

siwane:
This PR aims to delete `eventAggregator`.

After investigation, this service aggregate events and iterate on it to emit them. It also defer emission that is a breaking change, especially for `core_kernel_classes_Resource`


bartlomiejmarszal (approved):
LGTM

core/kernel/classes/class.Resource.php
```diff
@@ -764,8 +758,7 @@ protected function toResource($value)
 
     private function onUpdate(): void
     {
-        /** @var EventAggregator $eventAggregator */
-        $eventAggregator = $this->getServiceManager()->get(EventAggregator::SERVICE_ID);
-        $eventAggregator->put($this->getUri(), new ResourceUpdated($this));
+        $eventManager = $this->getServiceManager()->get(EventManager::SERVICE_ID);
+        $eventManager->trigger(new ResourceUpdated($this));
```
gabrielfs7, line 762: Now it executes the event, before It seems it just aggregated by URI, so when one URI is used, all the events are triggered. Will we have same behavior now? Why did we change it?

siwane: We have changed it because it is deferring event and we do not know what it will imply, it is a breaking changes

boajer (approved):

core/kernel/classes/class.Resource.php
gabrielfs7: After @andreluizmachado answer would we keep the removal?

siwane: I would say yes, he confirms that this service tries to improve the taskqueue mechanism and avoid multiple calls based on events. 
This workaround is good but the problem stay the same. I can close it, but number of triggered events will not changed (only the emition)
And, as discussed with @andreluizmachado , this fix is changing behavior of eventManager for the whole platform and can create side effect. Need to be confirmed by regression tests

PR #857, opened 2020-11-30
Feature/con 76/advance search in result page

bartlomiejmarszal:
Permission helper is now taking array key to recognise permissions for resources

https://oat-sa.atlassian.net/browse/CON-76

### DIFF
--- a/core/data/permission/PermissionHelper.php
+++ b/core/data/permission/PermissionHelper.php
@@ -38,9 +38,14 @@ class PermissionHelper extends ConfigurableService
         if (!in_array($right, $provider->getSupportedRights())) {
             return $resourceIds;
         }
-        $permissions = $provider->getPermissions($this->getCurrentUser(),$resourceIds);
+        $permissions = $provider->getPermissions($this->getCurrentUser(), array_keys($resourceIds));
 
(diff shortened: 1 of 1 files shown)


core/data/permission/PermissionHelper.php
```diff
@@ -38,9 +38,14 @@ public function filterByPermission(array $resourceIds, string $right): array
         if (!in_array($right, $provider->getSupportedRights())) {
             return $resourceIds;
         }
-        $permissions = $provider->getPermissions($this->getCurrentUser(),$resourceIds);
+        $permissions = $provider->getPermissions($this->getCurrentUser(), array_keys($resourceIds));
```
gabrielfs7, line 41: This is a tricky thing, we are assuming that the resourceIds now have some special key, which was not required before. This is dangerous can invisible by client of this class.

We should add this as another parameter in the method `filterByPermission` or pass objects instead of raw array.

Another option is to create a new method here with proper implementation

gabrielfs7 (changes requested):
Dangerous change causing invisible dependency.

habemuscode (approved):
COde review only

PR #859, opened 2021-01-07
Using stable version of symfony/polyfill-php80

gitromba:


Alroniks (approved):
Looks good

composer.json
```diff
@@ -67,7 +67,7 @@
     "monolog/monolog": "^1.23.0",
     "fluent/logger": "^1.0.1",
     "symfony/lock": "^3.4",
-    "symfony/polyfill-php80": "1.21.x-dev",
+    "symfony/polyfill-php80": "^1.20",
```
gabrielfs7, line 70: This new function will cause migration issues when moving to php8:

https://github.com/oat-sa/generis/commit/e4e843d6a0207654e2f59975070559f74afce977#diff-0c5f5d447957a7012b605c9fea0336c1f61017340a68d0797fa0d789be48f74cR165

@Alroniks Do you think we should stick to that?

gitromba:
I've replaced the `str_starts_with` calls with native implementation. Unit tests are passing. @gabrielfs7 @Alroniks FYI

gabrielfs7 (approved):
Code LGTM. I did a quick test locally and could not find any other usage of `str_starts_with` in our codebase.

Thanks @gitromba

PR #873, opened 2021-02-17
Laminas package name instead of outdated zend framework

alroniks:
The composer says that it is abandoned. It makes sense to update because sometime in the future it will be deleted and it make break our environments.


gitromba (approved):
Looking for a long time for this PR. Thanks a lot @alroniks 🙏

mike-ionut-mihai-sandu-tao:
guys i think this should not have been done, 
there are a lot of files where we use zend 
:: 
https://github.com/search?p=2&q=org%3Aoat-sa+zend%2FServiceManager&type=Code

PR #875, opened 2021-02-25
ci: fix vendor folder path

hutnikau:


gitromba (approved):
LGTM, thanks for the fix!

PR #878, opened 2021-02-26
Fix/legacy auto loader load from root folder

hutnikau:
Steps to reproduce:
1. clone tao-core repo: `git clone git@github.com:oat-sa/tao-core.git`
2. run composer update to install of dependencies of the package
3. require phpunit: `composer require "phpunit/phpunit:^8.5"` 
4. run unit tests: `vendor/bin/phpunit test/unit --bootstrap generis/test/bootstrap.php`

Expected results: tests are run
Actual results: tests run with `failed to open stream: No such file or director` error message

6. change generis dependency in composer.json: `"oat-sa/generis" : "dev-fix/LegacyAutoLoader_load_from_root_folder as 14.0.0",` and do `composer update`
7. run unit tests again

### DIFF
--- a/common/legacy/class.LegacyAutoLoader.php
+++ b/common/legacy/class.LegacyAutoLoader.php
@@ -19,6 +19,8 @@
  *
  */
 
+use oat\oatbox\extension\Manifest;
+
 /**
  * the generis autoloader
  *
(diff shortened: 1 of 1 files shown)


common/legacy/class.LegacyAutoLoader.php
```diff
@@ -129,6 +131,33 @@ public function autoload($pClassName)
...
+
+        if (file_exists($classFilePath)) {
+            require_once $classFilePath;
+            return;
+        }
+        if (file_exists($interfaceFilePath)) {
+            require_once $interfaceFilePath;
+            return;
         }
     }
```
gitromba, line 162: Considering the autoloading process is practically the same apart from the different path, I recommend the following simplifaction to avoid duplicating the logic:

```suggestion
    private function tryLoadFromRootFolder(string $extension, string $pClassName): void
    {
        if ($extension === (new Manifest($this->root . 'manifest.php'))->getName()) {
            $newClassName = preg_replace('/^[^_]+_/', '', $pClassName);
            // avoid being stuck in recursion
            ($newClassName !== $pClassName) && $this->autoload($newClassName);
        }
    }
```

To be called: 
```
$this->tryLoadFromRootFolder($token[0], $pClassName);
```

WDYT @hutnikau

hutnikau: Thanks for your suggestion. I would like to avoid recursion call there and try to stick to original implementation as close as possible. This is critical part of tao and any issue is there may cause big impact.

gitromba: Code duplication leads to issues thus my recommendation. At the end of course it's your call as being an expert here 😉

gitromba (approved):
Running `composer update` on this extension makes me able to successfully run its unit tests.

alroniks (approved):
It works for me and as a temporary solution, I agree. I hope we will find time to convert all this stuff to PSR-4.

PR #886, opened 2021-03-09
Fix/unit tests root path

hutnikau:
At time of launch of unit tests the `ROOT_PATH` constant is empty string what causes failures: http://qa.kitchen.it.taocloud.org/blue/organizations/jenkins/ci-pipeline-cg-extension-pr%2Ftao-core/detail/PR-2812/2/pipeline

The `ROOT_PATH` constant should be defined in accordance with the way how generis installed.

### DIFF
--- a/test/bootstrap.php
+++ b/test/bootstrap.php
@@ -21,15 +21,26 @@
  * @subpackage
  */
 
-$extensionRoot = __DIR__ . '/../';
+$extensionRoot = realpath(__DIR__ . '/../');
 
-// Use TAO dependency resolver first.
-if (is_dir($extensionRoot . '../vendor')) {
-    require_once $extensionRoot . '../vendor/autoload.php';
-} elseif (is_dir($extensionRoot . 'vendor')){
-    require_once $extensionRoot . 'vendor/autoload.php';
+function generisInstalledAsExtension(string $extensionRoot)
+{
+    return is_dir($extensionRoot . '/../vendor') && is_dir($extensionRoot . '/../generis');
+}
+
+function generisInstalledAsRootPackage(string $extensionRoot)
+{
+    return is_dir($extensionRoot . '/vendor');
+}
+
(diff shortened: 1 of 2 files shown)


test/bootstrap.php
```diff
@@ -21,15 +21,26 @@
...
+$extensionRoot = realpath(__DIR__ . '/../');
 
-// Use TAO dependency resolver first.
-if (is_dir($extensionRoot . '../vendor')) {
-    require_once $extensionRoot . '../vendor/autoload.php';
-} elseif (is_dir($extensionRoot . 'vendor')){
-    require_once $extensionRoot . 'vendor/autoload.php';
+function generisInstalledAsExtension(string $extensionRoot)
+{
+    return is_dir($extensionRoot . '/../vendor');
```
gitromba, line 28:
```suggestion
    return is_dir($extensionRoot . '/../vendor/generis');
```

test/config/generis.conf.php
```diff
@@ -39,7 +39,9 @@
 define('GENERIS_SESSION_NAME', 'test');
 
 # paths
-define('ROOT_PATH', '');
+if (!defined('ROOT_PATH')) {
+    define('ROOT_PATH', '');
```
gitromba, line 43: Are there entrypoints to this file other than `test/bootstrap.php`? 🤔

gitromba (changes requested):
Thanks @hutnikau. Please check out my comments 🙂

test/config/generis.conf.php
hutnikau: I do not know. I'm afraid that it may be used by other extensions so I left definition of this constant there.

gitromba: To guarantee proper test configuration the bootstrap file always need to be used. All these cases should be unveiled and fix if there's any rather than being suppressed IMO.

gitromba (approved):
LGTM

PR #910, opened 2021-08-18
Allow interface language to be overwritten in the BasicSession

quintanilhar:
## Ticket
https://oat-sa.atlassian.net/browse/TR-32

## Summary
This change allows the interface language to be overwritten on the session level so that we can change the language in the test runner according to the language set for the Delivery.

## Dependency
This PR relates to https://github.com/oat-sa/extension-tao-delivery/pull/511 and https://github.com/oat-sa/extension-tao-delivery-rdf/pull/391.

## How to test
See https://github.com/oat-sa/extension-tao-delivery/pull/511

### DIFF
--- a/common/session/class.BasicSession.php
+++ b/common/session/class.BasicSession.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
  *
  */
 
@@ -142,6 +142,10 @@ class common_session_BasicSession implements common_session_Session, ServiceLoca
      */
     public function getInterfaceLanguage()
     {
+        if (PHPSession::singleton()->hasAttribute('overrideInterfaceLanguage')) {
+            return PHPSession::singleton()->getAttribute('overrideInterfaceLanguage');
+        }
+
         return $this->getServiceLocator()->get(UserLanguageServiceInterface::class)->getInterfaceLanguage($this->getUser());
     }
     


common/session/class.BasicSession.php
```diff
@@ -142,6 +142,10 @@ public function getDataLanguage()
      */
     public function getInterfaceLanguage()
     {
+        if (PHPSession::singleton()->hasAttribute('overrideInterfaceLanguage')) {
```
alexeyvy, line 145: There's a wrapper around PHPSession that we could rely on to avoid make static calls so we further replace it easily:

`$this->getServiceLocator()->get(SessionService::SERVICE_ID)->hasAttribute('overrideInterfaceLanguage')
`
(from oat\oatbox\session\SessionService).

Also, what do you think about adding a new method like 'setInterfaceLanguage' to this interface instead? Are there blockers? Because I think it might be more obvious than reading a session attribute

wazelin: I might be wrong, but I don't think `SessionService` provides such a method.

common/session/class.BasicSession.php
```diff
@@ -142,6 +142,10 @@ public function getDataLanguage()
...
+            return PHPSession::singleton()->getAttribute('overrideInterfaceLanguage');
+        }
+
         return $this->getServiceLocator()->get(UserLanguageServiceInterface::class)->getInterfaceLanguage($this->getUser());
```
wazelin, line 149: WDYT of adding that overriding value check on the `UserLanguageServiceInterface` [implementation](https://github.com/oat-sa/generis/blob/master/common/oatbox/user/UserLanguageService.php#L57) level?

You can then easily extend [the test cases](https://github.com/oat-sa/generis/blob/master/test/unit/oatbox/user/UserLanguageServiceTest.php#L43) covering that method. You will also ensure the interface language is being overridden consistently, as some other pieces of code might be using the `UserLanguageService` directly.

wazelin (changes requested):

common/session/class.BasicSession.php
quintanilhar: Unfortunately, I can not use this service because it provides a limited session object. I could not find any other option to access PHPSession other than that. :(

common/session/class.BasicSession.php
quintanilhar: Good idea, I have isolated the logic into the UserLanguageService implementation and use the session to control when it should be set to avoid having a hard dependency with the session.

common/session/class.BasicSession.php
```diff
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
wazelin, line 18: Minor nitpick.

```suggestion
 * Copyright (c) 2017-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```

wazelin (approved):
- [x] New code is covered by tests (if applicable)
- [x] Feature is working correctly on my local machine (if applicable)

common/session/class.BasicSession.php
quintanilhar: Applied :)

alexeyvy (approved):
- [x] New code is covered by tests (if applicable)
- [x] Feature is working correctly on my local machine (if applicable)

PR #950, opened 2021-12-13
ADF-842: Make UserLanguageService::getDefaultLanguage() configurable

hectoras:
Associated Jira task: [ADF-842](https://oat-sa.atlassian.net/browse/ADF-842)

These changes allow configuring the laguage returned by  `UserLanguageService::getDefaultLanguage()` via conf files.

Related pull requests:
- https://github.com/oat-sa/extension-tao-udir/pull/142
- https://github.com/oat-sa/tao-core/pull/3269


common/oatbox/user/UserLanguageService.php
```diff
@@ -38,7 +38,11 @@ class UserLanguageService extends ConfigurableService implements UserLanguageSer
      */
     public function getDefaultLanguage()
     {
-        return DEFAULT_LANG;
+        if($this->hasOption(self::OPTION_DEFAULT_LANGUAGE)) {
+            $language = $this->getOption(self::OPTION_DEFAULT_LANGUAGE);
+        }
+
+        return $language ?? DEFAULT_LANG;
```
shpran, line 45:
```suggestion
        return $this->getOption(self::OPTION_DEFAULT_LANGUAGE, DEFAULT_LANG);
```

config/default/UserLanguageService.conf.php
```diff
@@ -4,4 +4,5 @@
 
 return new UserLanguageService([
     UserLanguageService::OPTION_LOCK_DATA_LANGUAGE => true,
+    UserLanguageService::OPTION_DEFAULT_LANGUAGE => 'nb-NO',
```
shpran, line 7: This will work for new instance (it's ok), but also you need to create a migration, which will override the current config with the new one.

common/oatbox/user/UserLanguageService.php
hectoras: Changed in fb83b41c70fce3419ef8b131f5fa30e6e0a806bb, thanks :)

config/default/UserLanguageService.conf.php
hectoras: I think this should work: 8de93a690bb8bf58f37f15c2340caec6509181b7. @shpran can you please take a quick look and check that is OK? :)

test/unit/oatbox/user/UserLanguageServiceTest.php
```diff
@@ -70,15 +110,15 @@ public function testGetDataLanguage()
 
     public function testIsDataLanguageEnabled()
     {
-        $service = $this->getService();
+        $service = new UserLanguageService();
```
gabrielfs7, line 113: Why we stoped using the previous method and now we are repeating the constructor several times?

common/oatbox/user/UserLanguageService.php
```diff
@@ -38,7 +38,7 @@ class UserLanguageService extends ConfigurableService implements UserLanguageSer
      */
     public function getDefaultLanguage()
     {
-        return DEFAULT_LANG;
+        return $this->getOption(self::OPTION_DEFAULT_LANGUAGE, DEFAULT_LANG);
```
ktsirkou, line 41: Be extra careful with this change. The method is by the next method `getDataLanguage`. 
So in UDIR case you will set data language to Norwegian and this will break all the the resources label rendering and the properties. 

Consider using the new default language definition only for the interface language.

test/unit/oatbox/user/UserLanguageServiceTest.php
hectoras: The previous method `getService()` just created and returned a new `UserLanguageService` instance with the provided options.

I found removing that indirection and instantiating the class unser test directly on each test case was a bit more clear (previously we called the `getService()` several times, now we directly call the constructor itself several times).

common/oatbox/user/UserLanguageService.php
gabrielfs7: We should only override the interface language in this case

test/unit/oatbox/user/UserLanguageServiceTest.php
gabrielfs7: I know, but if you have extra default parameters to the constructor os a default setter, etc you need to change in all places. 

I still find useful a factory method as it was before following DRY principle. It is a suggestion :)

common/oatbox/user/UserLanguageService.php
hectoras: I have rechecked how this work after the [comment in the tao-udir pull request](https://github.com/oat-sa/extension-tao-udir/pull/142#discussion_r771368566) and I've reached the same conclusion :sweat_smile: I'll change this to make the interface language configurable instead of this one.

test/unit/oatbox/user/UserLanguageServiceTest.php
hectoras: Okay, I'm restoring the previous method in 802e14877cb5ee2221a187526f4fb023eae5593c

common/oatbox/user/UserLanguageService.php
hectoras: Making `getInterfaceLanguage()` configurable, reverting `getDefaultLanguage()` to its former behavior, in 4dbf4834ab3fa856658c763625d7c74b7ccd77d8

common/oatbox/user/UserLanguageService.php
```diff
@@ -64,7 +64,11 @@ public function getInterfaceLanguage(User $user)
         }
 
         $lang = $user->getPropertyValues(GenerisRdf::PROPERTY_USER_UILG);
-        return empty($lang) ? $this->getDefaultLanguage() : (string)current($lang);
+        if(!empty($lang)) {
+            return (string)current($lang);
+        }
+
+        return $this->getOption(self::OPTION_DEFAULT_LANGUAGE, DEFAULT_LANG);
```
gabrielfs7, line 71: We generally add a blank line here. Can simplify return. Totally option, just cosmetics.
```suggestion

        return empty($lang) ? $this->getOption(self::OPTION_DEFAULT_LANGUAGE, DEFAULT_LANG) : (string)current($lang);
        }
```

config/default/UserLanguageService.conf.php
gabrielfs7: I am not sure we should call it `DEFAULT_INTERFACE_LANGUAGE` now 🤔

gabrielfs7, line 7: - [ ] Please revert this change. This is generis and should not be aware of this. 

```suggestion
```

hectoras: I'll rename it ro `OPTION_INTERFACE_LANGUAGE` / `interface_language`.

Would it be better to drop it altogether from this file (Generis) and just keep the migrations that sets its value to nb-NO in the client extension?

hectoras: Renamed as interface_language in bf796f4c79b697834f18acffdea978ffc350249d

core/kernel/users/class.GenerisUser.php
```diff
@@ -75,32 +75,53 @@ public function getPropertyValues($property)
...
+                    (string) $resource->getUniquePropertyValue(
+                        new core_kernel_classes_Property(OntologyRdf::RDF_VALUE)
+                    )
+                ];
+            } else {
+                common_Logger::w("Language {$resource} is not a resource");
+            }
+        }
+
+        return $language;
```
gabrielfs7, line 123: You can avoid `else` and make your code less complex, more readable. You can end the flow earlier for the reader.

https://williamdurand.fr/2013/06/03/object-calisthenics/#2-dont-use-the-else-keyword

```suggestion
        if (is_null($resource)) {
            return $language;
        }
        
        if (!$resource instanceof core_kernel_classes_Resource) {
            common_Logger::w("Language {$resource} is not a resource");
            
            return $language;
        }
        
       return [(string) $resource->getUniquePropertyValue(new core_kernel_classes_Property(OntologyRdf::RDF_VALUE))];
```

core/kernel/users/class.GenerisUser.php
```diff
@@ -75,32 +75,53 @@ public function getPropertyValues($property)
...
         return $this->cache[$property];
     }
 
     private function getUncached($property)
     {
-        $value = [];
         switch ($property) {
             case GenerisRdf::PROPERTY_USER_DEFLG:
+                $lang = $this->findLanguage($property);
+                return !empty($lang) ? $lang : [DEFAULT_LANG];
```
gabrielfs7, line 87: Are you sure we want to change it? Is it not just UI language in the scope?

config/default/UserLanguageService.conf.php
hectoras: :white_check_mark:  Already reverted: [UserLanguageService.conf.php](https://github.com/oat-sa/generis/blob/feature/ADF-842-configurable-user-default-language/config/default/UserLanguageService.conf.php)

core/kernel/users/class.GenerisUser.php
hectoras: :point_right: 20ae5d090399150f64f0c4019bd5be1399a97f8a :heavy_check_mark:

core/kernel/users/class.GenerisUser.php
hectoras: I've rechecked this, and I think it is correct as it is now. Still, is a bit tricky for me after all the recent changes, so let me explain my rationale:

Previously we handled both `PROPERTY_USER_DEFLG` and `PROPERTY_USER_UILG` the same way (that is, using the logic that is now in `findLanguage()` **and also** returning `DEFAULT_LANG` if the language was not set).  We moved that logic to a common method because it applies to both cases but now we are changing the return value on the error case for one of them.

`findLanguage()` does **not** return `DEFAULT_LANG` if `$property` is not set for the user, so what we are doing now is passing the return value from `findLanguage()` directly for `PROPERTY_USER_UILG` (so `UserLanguageService::getInterfaceLanguage()` [[here](https://github.com/oat-sa/generis/blob/01fe296d5c83561999a011bc6a230708d7323085/common/oatbox/user/UserLanguageService.php#L60)] receives it as `null` if not set).
- In that case, the caller (`UserLanguageService::getInterfaceLanguage()`) will pick its return return  value from `OPTION_INTERFACE_LANGUAGE`.

Now, for this case (`PROPERTY_USER_DEFLG`), `findLanguage()` behaves the same way (returns `null` if the language is not set) **but** we need to return the `DEFAULT_LANG` from here to maintain the old behavior.
- Therefore, we keep returning `DEFAULT_LANG` explicitly here for the "error" case of `findLanguage()` (i.e. when it returns `null`).

Do you think we are missing something?

gabrielfs7: Thanks for the explanation :) 

But aren't we addressing in this task is only the _interface language_? Why should we escalate this to the default language? Also, keep in mind that when @ktsirkou changed the default language for UDIR installation we had many problems parsing the RDF records and the UI broke.

gabrielfs7: You are right. The return value will be null and we will apply the change. Closing the thread :)

test/unit/oatbox/user/UserLanguageServiceTest.php
```diff
@@ -73,7 +115,7 @@ public function testIsDataLanguageEnabled()
         $service = $this->getService();
         $this->assertEquals(true, $service->isDataLanguageEnabled());
 
-        $service = $this->getService([
+        $service = new UserLanguageService([
```
gabrielfs7, line 118: You can use the previous method.

```suggestion
        $service = $this->getService([
```

common/oatbox/user/UserLanguageService.php
```diff
@@ -64,7 +64,8 @@ public function getInterfaceLanguage(User $user)
         }
 
         $lang = $user->getPropertyValues(GenerisRdf::PROPERTY_USER_UILG);
-        return empty($lang) ? $this->getDefaultLanguage() : (string)current($lang);
+
+        return empty($lang) ? $this->getOption(self::OPTION_INTERFACE_LANGUAGE, DEFAULT_LANG) : (string)current($lang);
```
gabrielfs7, line 68: - [ ] Please update unit tests

gabrielfs7 (approved):
Good job!

test/unit/oatbox/user/UserLanguageServiceTest.php
hectoras: I missed that one while rebasing with latest changes from develop, but I've already changed it again on 876a8d453f0b1f95ffd03db8ebc9305e855d8447 :ok_hand:

common/oatbox/user/UserLanguageService.php
hectoras: Here they are :point_right: [UserLanguageServiceTest.php](https://github.com/oat-sa/generis/blob/876a8d453f0b1f95ffd03db8ebc9305e855d8447/test/unit/oatbox/user/UserLanguageServiceTest.php)

They cover:
- User with UI lang, `OPTION_INTERFACE_LANGUAGE` set --> Returns user UI lang
- User with UI lang set, `OPTION_INTERFACE_LANGUAGE` **not** set --> Returns user UI lang
- User with **no** UI lang set, `OPTION_INTERFACE_LANGUAGE` set --> Returns `OPTION_INTERFACE_LANGUAGE`
- User with **no** UI lang set, `OPTION_INTERFACE_LANGUAGE` **not** set --> Returns `DEFAULT_LANG`

I've checked all branches from this method are reported as covered by PHPUnit if put this way:

```
public function getInterfaceLanguage(User $user)
{
    if (!empty($this->customInterfaceLanguage)) {
        return $this->customInterfaceLanguage;
    }

    $lang = $user->getPropertyValues(GenerisRdf::PROPERTY_USER_UILG);

    // return empty($lang) ? $this->getOption(self::OPTION_INTERFACE_LANGUAGE, DEFAULT_LANG) : (string)current($lang);

    if (empty($lang)) {
        return $this->getOption(self::OPTION_INTERFACE_LANGUAGE, DEFAULT_LANG);
    }

    return (string)current($lang);
}
```

core/kernel/users/class.GenerisUser.php
```diff
@@ -75,33 +75,56 @@ public function getPropertyValues($property)
...
-    
+
+    /**
+     * @param $property
+     * @return array|string[]
+     * @throws common_Exception
+     * @throws core_kernel_classes_EmptyProperty
+     * @throws core_kernel_persistence_Exception
+     */
+    private function findLanguage($property): array
```
shpran, line 104: Is it possible to declare `$property` type?

shpran (approved):
Good job! 👍🏻

common/oatbox/user/UserLanguageService.php
hectoras: This was restored to its former behavior (see  [here](https://github.com/oat-sa/generis/blob/feature/ADF-842-configurable-user-default-language/common/oatbox/user/UserLanguageService.php#L39)), marking this thread as solved :slightly_smiling_face:

test/unit/oatbox/user/UserLanguageServiceTest.php
hectoras: This was already restored ([here](https://github.com/oat-sa/generis/blob/feature/ADF-842-configurable-user-default-language/test/unit/oatbox/user/UserLanguageServiceTest.php#L149)), marking this thread as solved.

core/kernel/users/class.GenerisUser.php
hectoras: Sure, added here: cae3f3d354a9e63dbb1809dc31a357300e504abc :+1:

config/default/UserLanguageService.conf.php
hectoras: This .conf is not being modified by this pull request anymore (we are modifying it from client-specific migrations instead). Therefore I'm closing this thread.

common/oatbox/user/UserLanguageService.php
hectoras: I'm marking this thread as solved since all branches from this method are covered by unit tests.

core/kernel/users/class.GenerisUser.php
```diff
@@ -75,33 +75,55 @@ public function getPropertyValues($property)
...
         return $this->cache[$property];
     }
 
     private function getUncached($property)
     {
-        $value = [];
         switch ($property) {
             case GenerisRdf::PROPERTY_USER_DEFLG:
+                $lang = $this->findLanguage($property);
+                return !empty($lang) ? $lang : [DEFAULT_LANG];
```
shpran, line 87:
```suggestion
                return $lang ?: [DEFAULT_LANG];
```

shpran, line 87:
```suggestion
                $lang = $this->findLanguage($property);

                return $lang ?: [DEFAULT_LANG];
```

hectoras:
This pull request has 2 approvals and QA verified these changes. Therefore, I'm merging this pull request.

hectoras:
Released as [v15.13.1](https://github.com/oat-sa/generis/releases/tag/v15.13.1).

PR #977, opened 2022-02-10
Update EasyRdf library to the latest release

alroniks:
[REL-112: TAO CG EasyRDF PHP 7.4 Support](https://oat-sa.atlassian.net/browse/REL-112)

This pull request should be merged first and after it should be released. Other extensions should update their dependencies to require a new major release.

---

Support of PHP 7.4 in the library was added in the major release [1.0.0](https://github.com/easyrdf/easyrdf/releases/tag/1.0.0) which is broke backward compatibility. It also requires updating the major release of `generis` as well. I required the release [1.1.1](https://github.com/easyrdf/easyrdf/releases/tag/1.1.1) as it brings some internal improvements but does not break anything, so it is safe for usage.

### DIFF
--- a/common/legacy/ClassAliases.php
+++ b/common/legacy/ClassAliases.php
@@ -1,0 +1,12 @@
+<?php
+
+/**
+ * This file used for migrating extensions to the new version of EasyRdf library.
+ * After updating all extension this part of code can be removed.
+ */
+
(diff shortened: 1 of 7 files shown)


siwane (approved):
good job

PR #998, opened 2022-03-15
fix: avoid build the container before the application is installed

gabrielfs7:
Allow installation to run without building the container twice, so it can only be built after all extensions are installed

### DIFF
--- a/core/DependencyInjection/ContainerBuilder.php
+++ b/core/DependencyInjection/ContainerBuilder.php
@@ -53,12 +53,16 @@ class ContainerBuilder extends SymfonyContainerBuilder
     /** @var MiddlewareExtensionsMapper|null */
     private $middlewareExtensionsMapper;
 
+    /** @var string|null */
+    private $configPath;
+
     public function __construct(
         string $cachePath,
         ContainerInterface $legacyContainer,
         bool $isDebugEnabled = null,
         ContainerCache $cache = null,
-        MiddlewareExtensionsMapper $middlewareExtensionsMapper = null
+        MiddlewareExtensionsMapper $middlewareExtensionsMapper = null,
+        string $configPath = null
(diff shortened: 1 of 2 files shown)


core/DependencyInjection/ContainerBuilder.php
```diff
@@ -198,4 +207,9 @@ private function getExtensionsManager(): common_ext_ExtensionsManager
     {
         return $this->legacyContainer->get(common_ext_ExtensionsManager::SERVICE_ID);
     }
+
+    private function isApplicationInstalled(): bool
+    {
+        return file_exists($this->configPath . 'generis/installation.conf.php');
```
hectoras, line 213: Suggestion just in case the `configPath` is provided without a trailing slash, so we are sure the Generis dir and `confiigPath` are separated by a slash.

```suggestion
        return file_exists(($this->configPath !== null ? rtrim($this->configPath, '/') : '.') . '/generis/installation.conf.php');
```

hectoras (approved):

core/DependencyInjection/ContainerBuilder.php
```diff
@@ -72,6 +76,7 @@ public function __construct(
 
         parent::__construct();
         $this->middlewareExtensionsMapper = $middlewareExtensionsMapper ?? new MiddlewareExtensionsMapper();
+        $this->configPath = $configPath  === null ? (defined('CONFIG_PATH') ? CONFIG_PATH : null) : $configPath;
```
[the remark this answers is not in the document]
gabrielfs7: Done, thanks!

core/DependencyInjection/ContainerBuilder.php
gabrielfs7: Done, thanks!

PR #1010, opened 2022-05-23
[Fix] Convert ontology models to int

shpran:
# [ADF-1036](https://oat-sa.atlassian.net/browse/ADF-1036)

## Changes
* Added migration to convert ontology models to int:
  * `readable`
  * `writeable`
  * `addTo`

### DIFF
--- a/config/default/ontology.conf.php
+++ b/config/default/ontology.conf.php
@@ -3,13 +3,15 @@
 use core_kernel_persistence_smoothsql_SmoothModel as SmoothModel;
 use oat\generis\model\kernel\persistence\smoothsql\search\ComplexSearchService;
 
-return new \core_kernel_persistence_smoothsql_SmoothModel([
+return new SmoothModel([
     SmoothModel::OPTION_PERSISTENCE => 'default',
-    SmoothModel::OPTION_READABLE_MODELS =>
-        [SmoothModel::DEFAULT_WRITABLE_MODEL, SmoothModel::DEFAULT_READ_ONLY_MODEL],
-    SmoothModel::OPTION_WRITEABLE_MODELS =>
-        [SmoothModel::DEFAULT_WRITABLE_MODEL],
-    SmoothModel::OPTION_NEW_TRIPLE_MODEL =>
+    SmoothModel::OPTION_READABLE_MODELS => [
         SmoothModel::DEFAULT_WRITABLE_MODEL,
-    SmoothModel::OPTION_SEARCH_SERVICE => ComplexSearchService::SERVICE_ID
+        SmoothModel::DEFAULT_READ_ONLY_MODEL,
+    ],
+    SmoothModel::OPTION_WRITEABLE_MODELS => [
+        SmoothModel::DEFAULT_WRITABLE_MODEL,
+    ],
+    SmoothModel::OPTION_NEW_TRIPLE_MODEL => SmoothModel::DEFAULT_WRITABLE_MODEL,
+    SmoothModel::OPTION_SEARCH_SERVICE => ComplexSearchService::SERVICE_ID,
 ]);
--- a/migrations/Version202205232110572348_generis.php
+++ b/migrations/Version202205232110572348_generis.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace oat\generis\migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use oat\generis\model\data\Ontology;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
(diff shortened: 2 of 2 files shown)


migrations/Version202205232110572348_generis.php
```diff
@@ -0,0 +1,61 @@
...
+    public function up(Schema $schema): void
+    {
+        $serviceLocator = $this->getServiceLocator();
+
+        /** @var SmoothModel $ontology */
+        $ontology = $this->getServiceLocator()->get(Ontology::SERVICE_ID);
+
+        $this->fixModels($ontology, SmoothModel::OPTION_READABLE_MODELS);
+        $this->fixModels($ontology, SmoothModel::OPTION_WRITEABLE_MODELS);
+        $this->fixModel($ontology, SmoothModel::OPTION_NEW_TRIPLE_MODEL);
```
augustas, line 28: I would like that this migration makes the config to match https://github.com/oat-sa/generis/blob/master/config/default/ontology.conf.php#L8-L14

migrations/Version202205232110572348_generis.php
```diff
@@ -0,0 +1,61 @@
...
+
```
augustas, line 29: Also please remove the `SmoothModel::OPTION_CACHE_SERVICE` part like here: https://github.com/oat-sa/generis/commit/061126ad186952f759d309ec0d1731f6ae8266d0#diff-801705e555d09d2a1a4cfb1454765ff1570f123b566efb15ccff7e380232da04L16

augustas (dismissed):

migrations/Version202205232110572348_generis.php
hectoras: If we change the migration to write a hardcoded list (as the one from default/generis.conf.php), I think we would lose any additional values that may be set by extensions (for example, https://github.com/oat-sa/extension-tao-workspace/blob/master/scripts/install/SetupWrapper.php)

hectoras: There is also the possibility to keep the values already stored in the config (as this migration does) but add also the default values if not present (maybe something like `array_unique(array_merge($defaults, $current_values))`, but I suppose it may be cases of users removing the default writable models on purpose.

I would keep this migration as it is.

shpran: I cannot do this.
We have environments, which contains other values, and I don't want to reset them to default values.
@augustas 

image

augustas: Why you don't want to reset them to default? What is the purpose of those extra values?

migrations/Version202205232110572348_generis.php
augustas: The problem was that `config/generis/ontology.conf.php` was not updated along with code changes, as was discussed in the related ticket, so the scope is to make this config file up to date, including the removal of this `cache` part, investigating that it won't break anything also in same scope 🙂

migrations/Version202205232110572348_generis.php
shpran: That's the case - I do not know why we need these values
Resetting them to default might break something

augustas: If we will not clean up the code and config then in time it will be harder and harder to maintain it, code will rot and product will fail. So cleanup whenever possible, boy scout rule :) https://github.com/oat-sa/generis/pull/737 removes model table, it was already [a deprecated feature](https://github.com/oat-sa/generis/pull/731#pullrequestreview-359879262). So please, keep it only if you find code that actually needs it

config/default/ontology.conf.php
```diff
@@ -3,13 +3,15 @@
 use core_kernel_persistence_smoothsql_SmoothModel as SmoothModel;
 use oat\generis\model\kernel\persistence\smoothsql\search\ComplexSearchService;
 
-return new \core_kernel_persistence_smoothsql_SmoothModel([
+return new SmoothModel([
```
gabrielfs7, line 6: IMO it is better to not have the alias at all, since we can clearly see what class we are talking about.

```suggestion
return new core_kernel_persistence_smoothsql_SmoothModel([
```

migrations/Version202205232110572348_generis.php
gabrielfs7: @augustas there might be the case some customer persisted the `model_id` in the database using those additional values, so it is a risky action IMO, to just remove them.

The lesson from this is that we should not create config and customization for unnecessary things, but now that is done, we can only keep it backward compatible IMO.

shpran: Due to this requested changes, I cannot merge this PR 😢

PR #1013, opened 2022-06-17
fix: ensure cache directory exists

augustas:
When TAO instance is reacreated and `taoUpdate.php` gets called, any migration trying to access DI container will fail with an error of missing cache directory. This change ensures the directory is present  

Related to : https://oat-sa.atlassian.net/browse/REL-783
   
Companion PR :
 - [ ] https://github.com/oat-sa/tao-core/pull/3561

### DIFF
--- a/common/oatbox/cache/SetupFileCache.php
+++ b/common/oatbox/cache/SetupFileCache.php
@@ -29,10 +29,13 @@ class SetupFileCache extends ConfigurableService
 {
     public const PERSISTENCE = 'cache';
 
+    /**
+     * @throws Exception
+     */
     public function createDirectory($cachePath): void
     {
         if (is_dir($cachePath) && is_writable($cachePath)) {
-            return true;
+            return;
         }
 
         if (!@mkdir($cachePath, 0700, true)) {
--- a/common/oatbox/cache/SetupFileCache.php
+++ b/common/oatbox/cache/SetupFileCache.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA ;
+ */
+
+declare(strict_types=1);
+
+namespace oat\oatbox\cache;
+
+use oat\generis\persistence\PersistenceManager;
+use oat\oatbox\service\ConfigurableService;
+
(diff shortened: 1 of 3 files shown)


core/DependencyInjection/ContainerBuilder.php
```diff
@@ -227,4 +229,9 @@ private function isApplicationInstalled(): bool
     {
         return file_exists(rtrim((string)$this->configPath, '/') . '/generis/installation.conf.php');
     }
+
+    private function ensureCacheDirectoryIsCreated(): void
+    {
+        @mkdir($this->cachePath, 0700, true);
```
gabrielfs7, line 235: Could we check if directory exist or is writable and create it otherwise, instead of using the `@` operator which will mask permission issues, etc?

I think this is something that we should solve on DevOps side, cause for config/data we break the installation if they do not exist and are not writable. 

Wdyt?

augustas: Updated the code to check for directory beforehand.

Whether the solution should be on infrastructure or application side, at first I wanted to solve it on infra code, but I think the argument is valid that app should ensure the folder structure it expects. And it is reasonable for infra to assume that `data` folder can be wiped, any files that must persist are stored in specifically configured file-systems.

gabrielfs7: I suggest to add this as a pre-install/pre-update script, cause DI container should not be creating generis cache:

```
if (!mkdir(GENERIS_CACHE_PATH, 0700, true) && !is_dir(GENERIS_CACHE_PATH)) {
    throw new InvalidArgumentException(
        sprintf(
            'Could not create application cache',
            $this->cachePath
        )
    );
}
```

common/oatbox/cache/SetupFileCache.php
```diff
@@ -0,0 +1,54 @@
...
+    public const PERSISTENCE = 'cache';
+
+    public function createDirectory($cachePath): bool
+    {
+        if (is_dir($cachePath) && is_writable($cachePath)) {
+            return true;
+        }
+
+        return @mkdir($cachePath, 0700, true);
+    }
```
gabrielfs7, line 38: I would throw an exception here instead and not return a boolean, since we always have to handle this boolean outside and throw an exception anyway. 

The exception can also contain better details about what happened where a boolean does not.

https://github.com/oat-sa/tao-core/pull/3561/files#r911034082

```suggestion
    public function createDirectory($cachePath): void
    {
        if (is_dir($cachePath) && is_writable($cachePath)) {
            return true;
        }

        if (!@mkdir($cachePath, 0700, true)) {
            throw new Exception(sprintf( 'Could not create application cache at "%s"', $cachePath) );
        }
    }
```

common/oatbox/cache/SetupFileCache.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function createPersistence(): void
```
gabrielfs7, line 40: Are you sure this is the place to have this method? This is a `SetupFileCache` script.

augustas: Thats how tao cache works, a KeyValueCache with `phpfile` driven `cache` persistence, the folder itself isn't a prerequisite, since setting a value creates the necessary folder structure. I believe this part can live in `SetupFileCache` 🤔

gabrielfs7: Got it. It makes sense in this case. :) 

Could I do a small optional suggestion? use the prefix `setup` in the public methods to be consistent with the class name. Example: `setupPersistence` and `setupStorage` (instead of `createDirectory`).

Or ir possible, have just a single method `setup` that does all the work, but this is optional suggestion :)

gabrielfs7 (approved):

PR #1014, opened 2022-07-01
Release 15.20.3

augustas:
- _Fix_ [REL-783](https://oat-sa.atlassian.net/browse/REL-783)  Extract TAO cache setup

### DIFF
--- a/common/oatbox/cache/SetupFileCache.php
+++ b/common/oatbox/cache/SetupFileCache.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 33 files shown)


common/oatbox/cache/SetupFileCache.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @throws Exception
+     */
+    public function createDirectory($cachePath): void
+    {
+        if (is_dir($cachePath) && is_writable($cachePath)) {
+            return;
+        }
+
+        if (!@mkdir($cachePath, 0700, true)) {
```
yaraslau-kavaliou, line 41: Do we really need to suppress errors using '@'?

augustas: Not essential, but makes the code shorter. Warnings themselves are not helpful in this case, an exception is thrown and logged if directory creation fails

yaraslau-kavaliou: I propose to do like this in order to not lose error message:
```
if (!@mkdir($cachePath, 0700, true)) {
    $error = error_get_last();
    throw new Exception(sprintf( 'Could not create application cache at "%s". Error message: %s', $cachePath, $error['message'] ?? 'unknown') );
}
```

augustas: I updated it, thanks for suggestion

yaraslau-kavaliou (approved):

PR #1030, opened 2023-03-12
feat: add missing Redis methods

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-1328
https://oat-sa.atlassian.net/browse/ADF-1329

# Goal

Add missing Driver methods for Redis

## Related PRs

- https://github.com/oat-sa/tao-core/pull/3751

## TODO

- [ ] Add unit tests 
- [x] Fix code styles

### DIFF
--- a/common/persistence/class.PhpRedisDriver.php
+++ b/common/persistence/class.PhpRedisDriver.php
@@ -253,7 +253,7 @@ class common_persistence_PhpRedisDriver implements common_persistence_AdvKvDrive
      */
     public function mDel(array $keys)
     {
-        return $this->callWithRetry('del', $keys);
+        return $this->callWithRetry('del', [$keys]);
     }
 
     /**
--- a/common/persistence/class.PhpRedisDriver.php
+++ b/common/persistence/class.PhpRedisDriver.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -113,11 +114,11 @@ class common_persistence_PhpRedisDriver implements common_persistence_AdvKvDrive
             try {
                 $result = call_user_func_array([$this->connection , $method], $params);
                 $success = true;
-            } catch (\Exception $e) {
+            } catch (Exception $e) {
                 $lastException = $e;
-                \common_Logger::d('Redis  ' . $method . ' failed ' . $attempt . '/' . $retry . ' :  ' . $e->getMessage());
+                common_Logger::d('Redis  ' . $method . ' failed ' . $attempt . '/' . $retry . ' :  ' . $e->getMessage());
                 if ($e->getMessage() == 'Failed to AUTH connection' && isset($this->params['password'])) {
-                    \common_Logger::d('Authenticating Redis connection');
+                    common_Logger::d('Authenticating Redis connection');
                     $this->connection->auth($this->params['password']);
                 }
                 $delay = rand(self::RETRY_DELAY, self::RETRY_DELAY * 2);
@@ -148,18 +149,18 @@ class common_persistence_PhpRedisDriver implements common_persistence_AdvKvDrive
         };
         return $this->callWithRetry('set', [$key, $value, $options]);
     }
-    
+
     public function get($key)
     {
 
         return $this->callWithRetry('get', [$key]);
     }
-    
+
     public function exists($key)
     {
         return (bool)$this->callWithRetry('exists', [$key]);
     }
-    
+
     public function del($key)
     {
         return $this->callWithRetry('del', [$key]);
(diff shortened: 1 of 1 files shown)


common/persistence/class.PhpRedisDriver.php
```diff
@@ -189,30 +192,94 @@ public function hGet($key, $field)
...
         return $this->callWithRetry('incr', [$key]);
     }
+
     //Time complexity: O(1)
     public function decr($key)
     {
         return $this->callWithRetry('decr', [$key]);
     }
 
+    /**
```
rgomeztao, line 222: question: im not sure whats the scope of this PR :). i can see there are new functions, but shouldnt this functions be specified in a contract interface? and if so, shouldnt the other implementations include this functions to not break when used? (i can see this PR doesnt use this methods anywhere so its safe, just a bit lost :D)

common/persistence/class.PhpRedisDriver.php
```diff
@@ -189,30 +192,94 @@ public function hGet($key, $field)
...
+    public function scan(int &$iterator = null, string $pattern = null, int $count = 1000): array
+    {
+        $retry = $this->params['attempt'];
+        $attempt = 0;
+
+        while ($attempt <= $retry) {
+            try {
+                return $this->connection->scan($iterator, $pattern, $count);
+            } catch (Exception $exception) {
+                common_Logger::d('Redis  ' . $method . ' failed ' . $attempt . '/' . $retry . ' :  ' . $exception->getMessage());
```
rgomeztao, line 235: mmm sprintf could be better? :)

common/persistence/class.PhpRedisDriver.php
```diff
@@ -189,30 +192,94 @@ public function hGet($key, $field)
...
+
+                if ($e->getMessage() == 'Failed to AUTH connection' && isset($this->params['password'])) {
```
rgomeztao, line 237: 'Failed to AUTH connection' a constant for this would be nice :)

hectoras, line 237: It's weird to have a conditional based on the exception message, but as I see the exception code is always thrown as zero: https://github.com/phpredis/phpredis/blob/develop/redis.c#L604

rgomeztao: yeah :/ its pretty dirty

hectoras: I'd maybe add a comment saying why we compare against a exception message (i.e. because (I think) the extension always uses the same exception class and there is no exception code we can use to get why it failed).

common/persistence/class.PhpRedisDriver.php
gabrielfs7: Adding to the contract will cause a breaking change and I would have to implement them to all drivers. That is why I did not :)

common/persistence/class.PhpRedisDriver.php
gabrielfs7: Check the file. I copied the same logic inside `callWithRetry` (cause it does not work for `scan` method). I can make it better for sure :)

gabrielfs7: https://github.com/oat-sa/generis/pull/1030#discussion_r1135504891

common/persistence/class.PhpRedisDriver.php
```diff
@@ -103,33 +104,28 @@ private function connectToCluster(array $host, int $timeout, bool $persist)
...
         while (!$success && $attempt <= $retry) {
             try {
-                $result = call_user_func_array([$this->connection , $method], $params);
+                $result = call_user_func_array([$this->connection, $method], $params);
                 $success = true;
-            } catch (\Exception $e) {
+            } catch (Exception $e) {
                 $lastException = $e;
-                \common_Logger::d('Redis  ' . $method . ' failed ' . $attempt . '/' . $retry . ' :  ' . $e->getMessage());
-                if ($e->getMessage() == 'Failed to AUTH connection' && isset($this->params['password'])) {
```
gabrielfs7, line 119: This was existing logic for years, so I did not want to change the approach after the current small refactoring

hectoras (approved):
- [x] Target branch is correct
- [x] Commits are following conventional commits
- [x] Kitchen links are not exposed in the description
- [x] Update logs provided

rgomeztao (approved):
latest changes looks nice :), approved! :partying_face:

common/persistence/class.PhpRedisDriver.php
gabrielfs7: What is your suggestion? That is the problem with interfaces and open source projects, every change in the interface breaks change to everybody (customers and other users that we do not even know).

I am aware it is not the best approach, but do you have an idea to do it without breaking changing and under the budget we have (almost ending)?

PR #1033, opened 2023-03-16
chore: automate generis extension release

gabrielfs7:
# Goal 

Allow generis extension to be released automatically

# Why?

- We are spending to much time on releasing it manually
- There is no much benefit on it as many other core extensions also have automatic release

### DIFF
--- a/.github/workflows/release_tao_extension.yml
+++ b/.github/workflows/release_tao_extension.yml
@@ -1,0 +1,23 @@
+# Make an extension release after merging to develop branch
+
+name: Release Tao extension
+
+on:
+  pull_request:
+    branches:
+      - develop
+    types: [closed]
+jobs:
+  auto-release:
+    if: github.event.pull_request.merged == true
+    name: Automated Tao extension release
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Release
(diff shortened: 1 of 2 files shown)


wazelin (changes requested):
While we're at it, let's also add a workflow similar to https://github.com/oat-sa/tao-core/blob/master/.github/workflows/conventional_commits_check.yml. Having a release version preview in the PR might come in handy.

.github/workflows/release_tao_extension.yml
```diff
@@ -0,0 +1,23 @@
...
+    if: github.event.pull_request.merged == true
+    name: Automated Tao extension release
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Release
+        uses: oat-sa/extension-release-action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```
wazelin, line 21: That's no-longer required is it?

gabrielfs7: I am not sure, but tao-core and other extensions use it, so I am not sure we should remove.

wazelin: Fine by me. I see no evidence of that env var being used, but sure, it won't hurt keeping it.

wazelin (approved):

PR #1035, opened 2023-04-06
feat: database_index parameter for redis db selection

augustas:
Adds a new possible parameter for redis driver, `database_index`. The purpose is to select the redis database for the connection. This should make it possible to share a single Redis instance for several TAO applications.
 
Related to : https://oat-sa.atlassian.net/browse/REL-1037
 
#### How to test
 
- Setup several TAO instances 
- For all persistences that use `phpredis` driver, point to the same redis instance and specify a `database_index`, (integer 0 -> 15)
- Perform action in application that would store data in the persistences that have redis configured
- Check that only the specified redis databases have the data

https://user-images.githubusercontent.com/2705327/230311519-d57e3bfe-6c4e-4237-a411-e09d98a364a3.mp4

### DIFF
--- a/common/persistence/class.PhpRedisDriver.php
+++ b/common/persistence/class.PhpRedisDriver.php
@@ -86,10 +86,23 @@ class common_persistence_PhpRedisDriver implements common_persistence_AdvKvDrive
         } else {
             $this->connection->connect($host, $port, $timeout);
         }
+        if (isset($this->params['database_index'])) {
+            if (!$this->connection->select($this->params['database_index'])) {
+                $this->connection->close();
+                throw new common_exception_Error(
+                    "Failed to select Redis database"
+                );
+            }
+        }
     }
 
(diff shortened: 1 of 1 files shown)


vbyndych:
Your solution is really good @augustas and I don't see any issues why it will not work for us. I just want to propose an alternative solution so we could consider what will work beter in the end. But first please let me mention pros and cons of current solution, so compariso will be fair.

1. DB soultion
Pros:
 - data isolation on instance layer
 - no setup reuqired, works "out of the box".

Cons:
- limited number of spaces. We can olny handle 16 clients per instance.
- clustering is not supported.

2. Key namespacing via prefixing

Pros:
- unlimited number of namespaces per redis instance.
- supports any redis setup including clustering.

Cons:
- potentially less isolated, however with [Redis 7.0 ACL feature](https://redis.io/docs/management/security/acl/) we can achive same security level with just a bit more setup.

vbyndych (approved):

PR #1048, opened 2023-06-14
chore: add switch to follow redirects on request sent

uncleempty:
Ticket: https://oat-sa.atlassian.net/browse/LSI-2325

Adds a possibility to follow redirects when sending requests with legacy `Request` class via curl.
This is needed as a part of ISO-27001 changes. As an example: in the setup with SimpleRoster - we have problems after changing CloudFront origin settings to `Redirect HTTP to HTTPS`, as the `LtiOutcomeCallback` doesn't follow redirects and simply stops on getting 307 response code. This comes also from next:
all routes generated by SimpleRoster app are generated as HTTP routes (and the LtiOutcome is generated by SR). Ofc, we can force HTTPS scheme for SR routes, but this won't help because of point the way how we are working with SSL certs. We are using certs provided by AWS ACM. Those certificates cannot be exported, and due to this, the only way we can setup secure connections is apply the certificate on the load balancer side, not on the application side (via apache or other WS). And due  to this even if we are forcing the route to be HTTPS one, the HTTPS request would come only to load balancer. After - it is getting decrypted, and sent as a plain HTTP to the application instance. Which will return 404, since route is forced to be https.

So, this one, is atm the minimal changes need to make things working after applying ISO-27001-related security changes.

Small addition: this changes may look awkward, because not using `curl_setopt($handler, CURLOPT_FOLLOWLOCATION, 1);`, but this is not a solution, because in this case we are missing `Authorization` header, which also breaks the loop, since requests would be un-authorized on the SR side.

### DIFF
--- a/common/http/class.Request.php
+++ b/common/http/class.Request.php
@@ -33,6 +33,14 @@ class common_http_Request
 
     public const METHOD_GET = 'GET';
 
+    private const REDIRECT_CODES = [
+        301,
+        302,
+        303,
+        307,
+        308
+    ];
+
     /**
      * Creates an request from the current call
      *
@@ -188,12 +196,9 @@ class common_http_Request
     {
         return $this->body;
     }
-    /**
-     * @return common_http_Response
-     */
-    public function send()
-    {
 
+    public function send(bool $followRedirects = false): common_http_Response
+    {
         $curlHandler = curl_init($this->getUrl());
 
         //set the headers
@@ -244,8 +249,20 @@ class common_http_Request
         $httpResponse->headerOut = curl_getinfo($curlHandler, CURLINFO_HEADER_OUT);
         $httpResponse->effectiveUrl = curl_getinfo($curlHandler, CURLINFO_EFFECTIVE_URL);
         $httpResponse->responseData = $responseData;
+
+        $fullInfo = curl_getinfo($curlHandler);
+
         //curl_setopt($curlHandler, );
         curl_close($curlHandler);
+
+        if ($followRedirects && in_array($httpResponse->httpCode, self::REDIRECT_CODES, true)) {
+            $redirectUrl = $fullInfo['redirectUrl'] ?? '';
+            if ($redirectUrl) {
+                $this->url = $redirectUrl;
+                $httpResponse = $this->send();
+            }
+        }
+
         return $httpResponse;
     }
 
--- a/common/http/class.Request.php
+++ b/common/http/class.Request.php
@@ -242,6 +242,10 @@ class common_http_Request
         //curl_setopt($curlHandler, CURLINFO_HEADER_OUT, 1);
         //curl_setopt($curlHandler, CURLOPT_HEADER, true);
 
+        //directly setting `FOLLOWLOCATION` to false to make sure next lines would be working as expected
+        //and we can get the redirect url from curl
+        curl_setopt($curlHandler, CURLOPT_FOLLOWLOCATION, 0);
+
         $responseData = curl_exec($curlHandler);
         $httpResponse = new common_http_Response();
 
--- a/common/http/class.Request.php
+++ b/common/http/class.Request.php
@@ -255,13 +255,20 @@ class common_http_Request
         $httpResponse->responseData = $responseData;
 
         $redirectUrl = curl_getinfo($curlHandler, CURLINFO_REDIRECT_URL);
+        $sameDomain = null;
+        if ($redirectUrl) {
+            $initialDomain = parse_url($this->getUrl(), PHP_URL_HOST);
+            $redirectDomain = parse_url($redirectUrl, PHP_URL_HOST);
+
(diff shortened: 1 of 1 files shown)


common/http/class.Request.php
```diff
@@ -244,8 +249,20 @@ public function send()
         $httpResponse->headerOut = curl_getinfo($curlHandler, CURLINFO_HEADER_OUT);
         $httpResponse->effectiveUrl = curl_getinfo($curlHandler, CURLINFO_EFFECTIVE_URL);
         $httpResponse->responseData = $responseData;
+
+        $fullInfo = curl_getinfo($curlHandler);
+
         //curl_setopt($curlHandler, );
         curl_close($curlHandler);
+
+        if ($followRedirects && in_array($httpResponse->httpCode, self::REDIRECT_CODES, true)) {
```
wazelin, line 258: Don't you also need to set `CURLOPT_FOLLOWLOCATION` flag to 0 in order to disable curl-level redirects?

common/http/class.Request.php
```diff
@@ -237,15 +242,31 @@ public function send()
         //curl_setopt($curlHandler, CURLINFO_HEADER_OUT, 1);
         //curl_setopt($curlHandler, CURLOPT_HEADER, true);
 
+        //directly setting `FOLLOWLOCATION` to false to make sure next lines would be working as expected
+        //and we can get the redirect url from curl
+        curl_setopt($curlHandler, CURLOPT_FOLLOWLOCATION, 0);
```
wazelin, line 247: Also, a silly question.
Cannot we just do something like this instead of all that's added below?
```suggestion
        curl_setopt($curlHandler, CURLOPT_FOLLOWLOCATION, (int)$followRedirects);
```

common/http/class.Request.php
uncleempty: Copying also there the discussion in PM:
It shouldn't be set directly to `false`/`0` as it is either undeclared or already set to `false` in curl source.
However, I do fully agree with  the raised point, and to make sure the behaviour won't change - I'm adding this option too.
Thank you!

common/http/class.Request.php
```diff
@@ -237,15 +242,31 @@ public function send()
...
+
         $responseData = curl_exec($curlHandler);
         $httpResponse = new common_http_Response();
 
         $httpResponse->httpCode = curl_getinfo($curlHandler, CURLINFO_HTTP_CODE);
         $httpResponse->headerOut = curl_getinfo($curlHandler, CURLINFO_HEADER_OUT);
         $httpResponse->effectiveUrl = curl_getinfo($curlHandler, CURLINFO_EFFECTIVE_URL);
         $httpResponse->responseData = $responseData;
+
+        $fullInfo = curl_getinfo($curlHandler);
```
wazelin, line 257: Just a suggestion. You can add `CURLINFO_REDIRECT_URL` option here, and get the redirect URL value if any straight away. It will simplify the nested conditional statements below.

common/http/class.Request.php
uncleempty: Totally valid question.
Though, unfortunately, due to usage of `Authorization` header. If I correctly understood, from cURL 7.58.0 - it doesn't do so: https://curl.se/docs/CVE-2018-1000007.html

common/http/class.Request.php
```diff
@@ -244,8 +249,20 @@ public function send()
...
+
         //curl_setopt($curlHandler, );
         curl_close($curlHandler);
+
+        if ($followRedirects && in_array($httpResponse->httpCode, self::REDIRECT_CODES, true)) {
+            $redirectUrl = $fullInfo['redirectUrl'] ?? '';
+            if ($redirectUrl) {
+                $this->url = $redirectUrl;
+                $httpResponse = $this->send();
```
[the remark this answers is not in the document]
uncleempty: Good idea, though not fully sure if we really want to introduce this. At least without following the same direction as curl and adding second param to the `send` method and counting how many redirects already done. Just to make sure that we won't fall to any kind of redirect loop

wazelin (approved):

wazelin:

poyuki (approved):
LGTM
- [x] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

common/http/class.Request.php
wazelin: OK. One last suggestion.

WDYT about adding checks on the `$redirectUrl` before trying it?
Like, if it's still pointing to the same domain+path. I assume only the request schema should change from http to https.

wazelin (changes requested):

wazelin (approved):

uncleempty:
Agree with suggested change, though added only domain check, since redirect may be to the different endpoint in case of API changes. Added

PR #1059, opened 2023-08-10
fix: creating a directory for a service container

vbyndych:
## Goal
Make sure cache directory exists before creating a service container file. This way we can ensure no error happen during later steps.

## Changelog
- fix: creating a directory as a service container requires it.

## How to test / use
1. Install application
2. Remove `data` directory
3. Run the command `php index.php oat\generis\scripts\tools\ContainerCacheWarmup`

### DIFF
--- a/core/DependencyInjection/ContainerBuilder.php
+++ b/core/DependencyInjection/ContainerBuilder.php
@@ -101,6 +101,10 @@ class ContainerBuilder extends SymfonyContainerBuilder
             return $this->legacyContainer;
         }
 
+        if (!file_exists($this->cachePath)) {
+            mkdir($this->cachePath, 0700, true);
+        }
+
         if (!is_writable($this->cachePath)) {
             throw new InvalidArgumentException(
                 sprintf(


core/DependencyInjection/ContainerBuilder.php
```diff
@@ -101,6 +101,10 @@ public function forceBuild(): ContainerInterface
             return $this->legacyContainer;
         }
 
+        if (!file_exists($this->cachePath)) {
```
gabrielfs7, line 104: I think we are trying to solve an infra issue in the application.

An env without the cache dir created is the problem, so the **infra deployment should make sure it is created** with right permissions before deploy anything


⚠️ Delegating it to app can have other side effects, example, app user might not have permissions to create the directory or maybe parent directories to not exist, etc

So we should solve this on infra level as we do not have the same issue with exnterprise customers, so probably there the pipeline is correct.

gabrielfs7: If we sill want to create permissions here, we need to also validate if we have write permissions to create it and treat exception properly in case of no permissions, please :) 

Thanks!

gabrielfs7: As we discussed in out call we can use `oat\oatbox\cache\SetupFileCache` to create the directory in a previous step before calling container warmup command.

vbyndych: Yeah, thank you for the discussion. It seems like the problem with the pipeline has broader implications than I initially thought. That's why I think to create one more script similar to install or update which will take care of those things. Please check it out https://github.com/oat-sa/tao-core/pull/3870. I decided not to put it inside the cache warmer itself because it is not his responsibility and in case we would change the pipeline later, we would still be able to use the application cache warmer without any modifications. I'm closing this PR though as it is not needed anymore.

PR #1071, opened 2023-10-09
feat: rebuild and restart DI container in same request

augustas:
Helper function to both rebuild DI container and reset the cached object.
 
Related to : https://oat-sa.atlassian.net/browse/REL-1275
   
Companion PR :
 - [ ] https://github.com/oat-sa/tao-core/pull/3901


vbyndych (approved):
LGTM!

PR #1091, opened 2023-11-30
fix: Property label showed in schema changed form

yaraslau-kavaliou:
Ticket - https://oat-sa.atlassian.net/browse/REL-1370

Bug root cause:
* Having `string` types instead of  `core_kernel_classes_Literal` for literal values returning by `getPropertiesValues` method (in case of value itself is iterable)

How to test:
* Set up environment with neo4j persistence enabled
* Create a metadata property for item or test class
* Go to another page
* Return to the page with metadata editing
* Property label has to be set in form's label field
Screenshot from 2023-11-30 18-25-39

### DIFF
--- a/core/kernel/persistence/starsql/class.Resource.php
+++ b/core/kernel/persistence/starsql/class.Resource.php
@@ -543,22 +543,28 @@ CYPHER;
         foreach ($result->get('resource')->getProperties() as $key => $value) {
             if (in_array($key, $propertyUris)) {
                 if (is_iterable($value)) {
-                    $returnValue[$key] = array_merge(
-                        $returnValue[$key] ?? [],
-                        $this->getLanguageProcessor()->filterByLanguage($value, [$dataLanguage, $defaultLanguage])
-                    );
+                    $returnValue[$key] =
+                        array_map(
+                            fn($value) => $this->formatValue($value),
+                            array_merge(
+                                $returnValue[$key] ?? [],
+                                $this->getLanguageProcessor()->filterByLanguage(
+                                    $value,
+                                    [$dataLanguage, $defaultLanguage]
+                                )
+                            )
+                        );
                 } else {
-                    $returnValue[$key][] = common_Utils::isUri($value)
-                        ? $this->getModel()->getResource($value)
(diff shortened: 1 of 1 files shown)


core/kernel/persistence/starsql/class.Resource.php
```diff
@@ -543,22 +543,28 @@ public function getPropertiesValues(core_kernel_classes_Resource $resource, $pro
...
+                                )
+                            )
+                        );
                 } else {
-                    $returnValue[$key][] = common_Utils::isUri($value)
-                        ? $this->getModel()->getResource($value)
-                        : new core_kernel_classes_Literal($this->getLanguageProcessor()->parseTranslatedValue($value));
+                    $returnValue[$key][] = $this->formatValue(
+                        $value,
+                        [$this->getLanguageProcessor(), 'parseTranslatedValue']
```
vbyndych, line 560: **question**: do we need to check if a specific property is language-dependent before trying to apply regexp to trim it? I'm wondering as the regexp pattern could be costly on big chunks of text. Of course, it is a tradeoff here as calling check for language dependency will also cost something.

core/kernel/persistence/starsql/class.Resource.php
```diff
@@ -543,22 +543,28 @@ public function getPropertiesValues(core_kernel_classes_Resource $resource, $pro
         foreach ($result->get('resource')->getProperties() as $key => $value) {
             if (in_array($key, $propertyUris)) {
                 if (is_iterable($value)) {
-                    $returnValue[$key] = array_merge(
-                        $returnValue[$key] ?? [],
-                        $this->getLanguageProcessor()->filterByLanguage($value, [$dataLanguage, $defaultLanguage])
-                    );
+                    $returnValue[$key] =
```
vbyndych, line 546: **nitpick**: could you consider refactoring this part (extracting method refactoring in particular) as it becomes a bit difficult to read?

vbyndych (approved):
@yaraslau-kavaliou good work catching and fixing this issue. In general no anything blocking just a few things for a better look and feel.

PR #1095, opened 2023-12-12
feat: decrease query count by cacheing some RDF property fields

mccar:


pribi (approved):
LGTM, tested in local devel environment.

helpers/PropertyCache.php
```diff
@@ -2,7 +2,9 @@
 
 namespace oat\generis\Helper;
 
-class PropertyCache
+use Iterator;
+
+class PropertyCaches
```
pribi, line 7: Why did you change PropertyCache to PropertyCaches?

pribi: Fix: https://github.com/oat-sa/generis/pull/1097/files

PR #1102, opened 2024-01-10
REL-1315: NEO4J Inability to add Assets to Items as far as they are not presented in the stimuli bank

vbyndych:
## Goal
Fixed the following problem: "When the user tries to add any stimulus via CKEditor, the media bank presented empty, no matter if there are assets created"

## Changelog
- fix: escaped special characters which could be interpreted incorrectly within regex.

## How to test / use
1. Go to the Items tab
2. Create new item
3. Add any interaction to the field
4. Focus on the prompt/choice or any area available to be modified via CKE
5. Click “Insert Shared Stimulus“

After the fix:
Screenshot_2024-01-10_18-46-40


mike-ionut-mihai-sandu-tao (approved):
lgtm,
i did not test so let's make sure that we have a test in place for this

mccar (approved):

augustas (approved):

PR #1114, opened 2024-04-24
feat: Update extension to send QTI package

mccar:


common/class.Utils.php
```diff
@@ -207,4 +207,9 @@ public static function toHumanReadablePhpString($value, $indentation = 0)
 
         return (string) $returnValue;
     }
+
+    public static function getUnitId(string $uri): string
```
vbyndych, line 211: **suggestion**: I would suggest using a more generic name like resource ID or something like that because UnitId implies it is for tests (units).

PR #1120, opened 2024-06-10
chore: Add interface mode to rdf

Nevermind23:
## Task
https://oat-sa.atlassian.net/browse/AUT-3625

## Changelog
- Added simple interface mode in rdf

### DIFF
--- a/core/ontology/generis.rdf
+++ b/core/ontology/generis.rdf
@@ -381,4 +381,30 @@
     <rdfs:domain rdf:resource="http://www.w3.org/1999/02/22-rdf-syntax-ns#Property"/>
     <rdfs:range rdf:resource="http://www.w3.org/2000/01/rdf-schema#Property"/>
   </rdf:Description>
+
+  <rdf:Description rdf:about="http://www.tao.lu/Ontologies/generis.rdf#InterfaceModes">
+      <rdfs:subClassOf rdf:resource="http://www.tao.lu/Ontologies/TAO.rdf#SystemObject"/>
+      <rdfs:label xml:lang="en-US"><![CDATA[Interface Modes]]></rdfs:label>
+    </rdf:Description>
+    <rdf:Description rdf:about="http://www.tao.lu/Ontologies/generis.rdf#InterfaceMode">
+      <rdf:type rdf:resource="http://www.w3.org/1999/02/22-rdf-syntax-ns#Property"/>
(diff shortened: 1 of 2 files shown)


core/ontology/generis.rdf
```diff
@@ -381,4 +381,30 @@
     <rdfs:domain rdf:resource="http://www.w3.org/1999/02/22-rdf-syntax-ns#Property"/>
     <rdfs:range rdf:resource="http://www.w3.org/2000/01/rdf-schema#Property"/>
   </rdf:Description>
+
+  <rdf:Description rdf:about="http://www.tao.lu/Ontologies/generis.rdf#InterfaceModes">
+      <rdfs:subClassOf rdf:resource="http://www.tao.lu/Ontologies/TAO.rdf#SystemObject"/>
```
shpran, line 386: We need to think about using a different resource as parent, because this one relates to `TAO.rdf`, which belongs to child repository (generis -> tao).

shpran (changes requested):
Good job!
Please, add a migration to synchronize ontology changes during tao update.
Example of required migration - [click](https://github.com/oat-sa/tao-core/blob/72e99b88d0a9959064825b31d2512e248c4735d4/migrations/Version202110130633112234_tao.php#L22)

# How to "migrate"
* Create migration:
  ```
  php index.php 'oat\tao\scripts\tools\Migrations' -c generate -e [extensionsName]
  ```
* Execute all migrations:
  ```
  php tao/scripts/taoUpdate.php
  ```
* Execute specific migration:
  ```
  php index.php 'oat\tao\scripts\tools\Migrations' -c execute -v '<namespace>\<version>'
  ```
* Rollback specific migration:
  ```
  php index.php 'oat\tao\scripts\tools\Migrations' -c rollback -v '<namespace>\<version>'
  ```

core/ontology/generis.rdf
gabrielfs7: @Nevermind23 we probably need to use `http://www.tao.lu/datatypes/WidgetDefinitions.rdf#RadioBox`

gabrielfs7 (changes requested):
Thanks for the work @Nevermind23, but we still need to: 

- Use proper Widget to define radio button list
- Move this to tao extension as this does not belong to generis

gabrielfs7 (approved):

PR #1139, opened 2024-11-14
fix: add a fix to cover infinity recursive call

shpran:
## Goal
Use `getParentClassesIds` recursive call in `isInstanceOf` method to improve performance.

## Issue
When using `isInstanceOf` with `getParentClassesIds` recursive call on properties, it stuck due to infinity recursive call, because `http://www.w3.org/2000/01/rdf-schema#Class` type points to itself.

## TODO
- [ ] Check other types of resources to be compatible with new solution


gabrielfs7 (approved):
LGTM. 

Thanks @shpran

Before merge we need to ask QA

PR #1154, opened 2025-06-12
fix: move updating updated_at to the generis during global onUpdate

tikhanovichA:
Depends on https://github.com/oat-sa/tao-core/pull/4260

We found that during updating we also try to update the updatedAt which will also trigger onUpdate event which will be collected by the Event Aggregator
https://github.com/oat-sa/tao-core/blob/v54.39.2/models/classes/resources/ResourceWatcher.php#L95

For fixing, we will move updating the updatedAt inside onUpdate to prevent creating multi update events


Karol-Stelmaczonek (approved):

KirylHatalski (approved):

tikhanovichA:
Found more proper way to handle it in scope only tao-core

PR #1183, opened 2026-09-16
fix: INV-1278 avoid getUri() on false when deleting typeless resources

tikhanovichA:
## Summary
- After a successful ontology delete, `ResourceRepository` built `ResourceDeleted` with `$resourceType->getUri()` even when `getTypes()` was empty (`reset()` → `false`).
- That threw after the resource was already removed, so the UI reported failure while a reload showed the class gone.
- Guard empty types and set an empty `resourceType` string instead.
## Context
- New classes from `createSubClass()` only get `rdfs:subClassOf`, not `rdf:type`.
- Reproduced when deleting a class folder via the resource-selector tree (class delete incorrectly hit `deleteResource`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource deletion handling for resources without explicitly defined types.
  * Deletion events now record the resource type safely when type information is available.

## How to test
- Activate a new tree view by adding next config into `config/tao/client_lib_config_registry.conf.php`
` 'layout/tree/loader' => array(
    'treeProvider' => 'resource-selector'
),`

After the fix: 

https://github.com/user-attachments/assets/f9beede3-d974-45ea-8feb-7590f16ea7dc


core/resource/Repository/ResourceRepository.php
```diff
@@ -86,12 +86,17 @@ public function delete(ContextInterface $context): void
...
+        // subclasses often have rdfs:subClassOf only — getTypes() is empty
         $resourceDeletedEvent = (new ResourceDeleted($resource->getUri()))
             ->setSelectedClass($selectedClass)
             ->setParentClass($parentClass)
-            ->setResourceType($resourceType->getUri());
+            ->setResourceType(
+                $resourceType instanceof core_kernel_classes_Resource
+                    ? $resourceType->getUri()
+                    : ''
+            );
```
Karol-Stelmaczonek, line 99: Do we want to keep it only for `core_kernel_classes_Resource` or we can simply go with 
```suggestion
            ->setResourceType($resourceType?->getUri() ?? '');
```
?

Karol-Stelmaczonek (approved):

core/resource/Repository/ResourceRepository.php
tikhanovichA: but its tree only for core_kernel_classes_Resource

tikhanovichA: $resourceType comes from reset($resource->getTypes()), and on an empty list that returns false, not null. ?-> only guards null, so $resourceType?->getUri() still fatals on the empty-types case this change is meant to cover.
I’d keep the instanceof check

